PR #5, opened 2019-01-16
Add New Client To Proxy User Request

willmichael:
This is to add a new method that proxies user calls to PDS for the Executer it takes a bearer token and manually constructs the authorization header for the request.

Companion PR to WASH-501 on e3dbsearchservice, waiting on merge of WASH-500 to make PR for that.

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,27 +36,17 @@ func (err *HTTPError) Error() string {
 // MakeE3DBServiceCall attempts to call an e3db service by executing the provided request and deserializing the response into the provided result holder, returning error (if any).
 func MakeE3DBServiceCall(httpAuthorizer E3DBHTTPAuthorizer, ctx context.Context, request *http.Request, result interface{}) error {
 	client := httpAuthorizer.AuthHTTPClient(ctx)
-	response, err := client.Do(request)
-	if err != nil {
-		return err
-	}
-	defer response.Body.Close()
-	if !(response.StatusCode >= 200 && response.StatusCode <= 299) {
-		requestURL := request.URL.String()
-		return &HTTPError{
-			StatusCode: response.StatusCode,
-			URL:        requestURL,
-			message:    fmt.Sprintf("e3db: %s: server http error %d", requestURL, response.StatusCode),
-		}
-	}
-	err = json.NewDecoder(response.Body).Decode(&result)
-	return err
+	return makeServiceCall(client, request, result)
 }
 
 // MakeProxiedUserCall attempts to call an e3db service using provided user auth token to authenticate request.
 func MakeProxiedUserCall(ctx context.Context, userAuthToken string, request *http.Request, result interface{}) error {
 	client := &http.Client{}
 	request.Header.Add("Authorization", "Bearer "+userAuthToken)
+	return makeServiceCall(client, request, result)
+}
+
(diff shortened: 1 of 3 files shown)


pdsClient/pdsClient.go
```diff
@@ -1,110 +1,110 @@
...
-    if err != nil {
-        return result, err
-    }
-    request, err := http.NewRequest("POST", c.Host+"/"+PDSServiceBasePath+"/search", &buf)
-    if err != nil {
-        return result, err
-    }
-    err = e3dbClients.MakeE3DBServiceCall(c.E3DBAuthClient, ctx, request, &result)
-    return result, err
+func (c *E3dbPDSClient) ListRecords(ctx context.Context, authToken string, params ListRecordsRequest) (*ListRecordsResult, error) {
```
galxy25: This should be a separate method, we want to have one method that "normal" clients can use and one for proxy clients(like the executor) to use.

main.go
```diff
@@ -1,54 +1,75 @@
...
-        requestURL := request.URL.String()
-        return &HTTPError{
-            StatusCode: response.StatusCode,
-            URL:        requestURL,
-            message:    fmt.Sprintf("e3db: %s: server http error %d", requestURL, response.StatusCode),
-        }
-    }
-    err = json.NewDecoder(response.Body).Decode(&result)
-    return err
+	client := httpAuthorizer.AuthHTTPClient(ctx)
```
lkwdwrd: So this and the next function look very nearly identical. Is there a reason not to combine the common code?

pdsClient/pdsClient.go
galxy25: Do the tests still pass with this update as is?

willmichael: It does not (my mistake I did not commit updated test), but I have separated the methods and added line to test both methods.

galxy25: Awesome!

main.go
willmichael: They have distinctly different purposes so I like having them in different functions, but I could probably make a helper method to take out some of the repeated code as I just did with other methods in pdsClient. In this case I'm not sure if I see value, but maybe you see value in refactoring the two methods into one//adding a helper @galxy25?

pdsClient/pdsClient.go
```diff
@@ -1,110 +1,113 @@
...
+	if err != nil {
+		return result, err
+	}
+	err = e3dbClients.MakeE3DBServiceCall(c.E3DBAuthClient, ctx, request, &result)
+	return result, err
+}
+
+// createListRecordsRequest isolates duplicate code in creating http search request.
+func createRequest(method string, path string, params interface{}) (*http.Request, error) {
+	var buf bytes.Buffer
```
willmichael: Do you think this refactor obfuscates the functions it takes code out of? @galxy25

main.go
galxy25: Definitely two methods as they will be used for distinctly  different purposes, but there would be value in extracting the common code into a helper.

lkwdwrd: Sure -- helper seems like the right call here. The signatures of these functions are different, it just seems like this request pattern works exactly the same when it is used so may be worth abstracting. 👍

PR #6, opened 2019-01-21
Add go client for talking to search service indexer.

galxy25:
- Tests

### DIFF
--- a/searchIndexerClient/indexerClient.go
+++ b/searchIndexerClient/indexerClient.go
@@ -1,0 +1,49 @@
+package searchIndexerClient
+
+import (
+    "bytes"
+    "context"
+    "encoding/json"
+    "github.com/tozny/e3db-clients-go"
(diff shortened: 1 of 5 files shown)


searchIndexerClient/indexerClient.go
```diff
@@ -0,0 +1,49 @@
...
+
+//E3dbSearchIndexerClient implements an http client for communication with an e3db Search Indexer service.
+type E3dbSearchIndexerClient struct {
+    APIKey    string
+    APISecret string
+    Host      string
+    *authClient.E3dbAuthClient
+}
+
+// IndexRecord uses an internal(available only to locally running e3db instances) endpoint to register a client, returning the registered client and error (if any).
```
willmichael: IndexRecord registers a client?

galxy25: Good catch, copy pasta left over from `pdsClient.go`

PR #9, opened 2019-02-13
Add e3db go client methods for sharing records between clients and ge…

galxy25:
- Add e3db go client methods for sharing records between clients and getting access keys.
- Tests.
# TODO
- Convert all naming of FieldId's to FieldID

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/tozny/e3db-clients-go
 
 require (
+	github.com/google/uuid v1.1.0
 	github.com/stretchr/testify v1.3.0 // indirect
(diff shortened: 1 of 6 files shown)


galxy25:
@lkwdwrd this is not an sdk for external consumption, its for consolidating clients used by Tozny e3db-go services to talk to other e3db services and making it easier to write integration tests.

lkwdwrd:
Ah! OK, so this is basically a testing tool in the internal facing code. I don't think that's the first time I've gotten that confused. Carry on 👍

galxy25:
> does TestSharedRecordsCanBeFetchedBySharee pass?

@willmichael yes.

PR #10, opened 2019-02-14
Feature/delete indexed records

willmichael:
Add call to delete indexed records for use in integration tests by the search service.

### DIFF
--- a/pdsClient/api.go
+++ b/pdsClient/api.go
@@ -74,6 +74,11 @@ type WriteRecordRequest = Record
 // WriteRecordResponse represents a response from trying to write a record via the e3db api.
 type WriteRecordResponse = Record
 
+// DeleteRecordRequest represents a request to delete a record via e3db api.
+type DeleteRecordRequest struct {
+	RecordID string `json:"record_id"`
+}
+
 // ListRecordsRequest represents a valid ListRecord call to the PDS service.
 type ListRecordsRequest struct {
 	Count       int  `json:"count,omitempty"`        //The maximum number of records to return.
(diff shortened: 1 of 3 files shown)


pdsClient/pdsClient.go
```diff
@@ -139,6 +139,17 @@ func (c *E3dbPDSClient) WriteRecord(ctx context.Context, params WriteRecordReque
 	return result, err
 }
 
+// DeleteRecord attempts to delete a record in e3db, returning error (if any).
+func (c *E3dbPDSClient) DeleteRecord(ctx context.Context, params DeleteRecordRequest) error {
```
galxy25: Please write an integration test for this. You have to run the pds and search services, and export some env variables for the integration tests to talk to those endpoints(all documented in the readme, but to save you some time
```
export E3DB_API_URL=https://local.e3db.com
export E3DB_API_KEY_ID=1d66c182779e47714cd957d4760a121580598d68e064c948e5887662c197538b
export E3DB_API_KEY_SECRET=aaadbd1cc72870e6bd41befa1273b4513d0d1fa04fb913257eec6158b6a35d75
export E3DB_CLIENT_ID=864e4b87-9eda-43fb-ae6d-d07d4275c73a
```

pdsClient/api.go
```diff
@@ -74,6 +74,11 @@ type WriteRecordRequest = Record
 // WriteRecordResponse represents a response from trying to write a record via the e3db api.
 type WriteRecordResponse = Record
 
+// DeleteRecordRequest represents a request to delete a record via e3db api.
+type DeleteRecordRequest struct {
+	RecordID string `json:"record_ids"`
```
galxy25: `json:"record_id"` vs ids?

willmichael: good catch, In the swagger spec it is actually `record_ids`, but I don't serialize it here so I'll switch

PR #15, opened 2019-03-11
Add pdsClient method for adding authorizer

galxy25:
Add pdsClient method for authorizing a client to share records with other clients, integration tests.

### DIFF
--- a/pdsClient/pdsClient.go
+++ b/pdsClient/pdsClient.go
@@ -21,9 +21,15 @@ type E3dbPDSClient struct {
 }
 
 // readPolicy wraps an e3db API object
-// for read access to records
+// for granting read access to records
 type readPolicy struct {
 	Read map[string]interface{} `json:"read"`
+}
+
+// authorizePolicy wraps an e3db API object
+// for granting share access for records of a given type
+type authorizePolicy struct {
+	Authorize map[string]interface{} `json:"authorizer"`
 }
 
 // allowReadPolicy wraps an e3db API object granting read access to records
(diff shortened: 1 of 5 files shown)


pdsClient/pdsClient.go
```diff
@@ -192,6 +236,76 @@ func (c *E3dbPDSClient) InternalGetRecord(ctx context.Context, recordID string)
...
+		UserID:     params.UserID,
+		ReaderID:   params.UserID,
+		RecordType: params.RecordType,
+	})
+	if err != nil {
+		return err
+	}
+	encryptedAccessKey := getAccessKeyResponse.EAK
+	if encryptedAccessKey == "" {
+		return errors.New(fmt.Sprintf("no access key exists for records of type %s", params.RecordType))
```
willmichael: Is this method a helper method for tests?

In the python SDK, if no EAK exists it is created and put for both the sharer and sharee to allow sharing of record types that don't yet exist

pdsClient/pdsClient.go
```diff
@@ -192,6 +236,76 @@ func (c *E3dbPDSClient) InternalGetRecord(ctx context.Context, recordID string)
...
+	}
+	// TODO: Decrypt this key
+	// TODO: using the decrypted version of
+	// encryptedAccessKey and the public key of
+	// the reader specified in params,
+	// create an encrypted access key that
+	// the reader can decrypt
+	// Put the encrypted access key for the authorizer
```
willmichael: `// Put the encrypted access key for the reader`?

pdsClient/pdsClient.go
galxy25: Yes, and tests have a MakeClientWriterForRecord that creates an access key for the sharer/writer of a record.

pdsClient/pdsClient.go
galxy25: Good catch

PR #17, opened 2019-03-27
Feature/generic account calls

lkwdwrd:
generic account calls to allow additional calls in services as needs expand. Needs tests, but gets things started.

### DIFF
--- a/accountClient/accountClient.go
+++ b/accountClient/accountClient.go
@@ -20,31 +20,37 @@ type E3dbAccountClient struct {
 }
 
 // CreateAccount attempts to create an e3db account using the provided params, returning created account and error (if any).
-func (c *E3dbAccountClient) CreateAccount(ctx context.Context, params CreateAccountRequest) (*CreateAccountResponse, error) {
+func (c *E3dbAccountClient) CreateAccount(ctx context.Context, params CreateAccountRequest) (*CreateAccountResponse, *e3dbClients.RequestError) {
 	var result *CreateAccountResponse
 	path := c.Host + "/" + AccountServiceBasePath + "/profile"
 	request, err := e3dbClients.CreateRequest("POST", path, params)
 	if err != nil {
-		return result, err
+		return result, &e3dbClients.RequestError {
+			URL: path,
+			message: err.Error()
+		}
 	}
 	err = e3dbClients.MakeRawServiceCall(&http.Client{}, request, &result)
 	return result, err
 }
 
 // InternalGetClientAccount attempts to get the account id and other account information for the specified client id
-func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, clientID string) (*InternalGetClientAccountResponse, error) {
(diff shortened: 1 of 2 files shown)


accountClient/accountClient.go
```diff
@@ -20,29 +20,51 @@ type E3dbAccountClient struct {
...
+	request, err := e3dbClients.CreateRequest(method, internalPath, params)
+	if err != nil {
+		return err
+	}
+	err = e3dbClients.MakeE3DBServiceCall(c.E3dbAuthClient, ctx, request, &result)
+	return err
+}
+
+// InternalServiceCall will make a call to a path based on the internal service root, using the method and params sent.
+func (c *E3dbAccountClient) InternalServiceCall(ctx context.Context, path, method string, params interface{}, result interface{}) *e3dbClients.RequestError {
```
galxy25: I have a few issue's with this pattern. 
1.) It's not at the same level of abstraction as the other clients and their current methods which provide methods for concrete actions(writing a record, sharing records, getting an access key, etc...).  
2.) These methods are generic enough to be in main.go of this package 
3.) This is hiding the fact that the account service has a separate authorization pattern for getting request tokens and isn't saving future users of these methods the boilerplate of setting that up. 

All that said, I understand if these are quick and dirty implementations to help prove out the concept and hope that they involve into refined and tested methods that can be used by other repo's!

main.go
```diff
@@ -22,40 +22,49 @@ type E3DBHTTPAuthorizer interface {
...
 
-// Error implements the error interface for HTTPError.
-func (err *HTTPError) Error() string {
+// Error implements the error interface for RequestError.
+func (err *RequestError) Error() string {
 	return err.message
 }
 
+// NewError creates a new RequestError
+func NewError(message, url string, statusCode int) *RequestError {
```
willmichael: small: `NewRequestError` in case we want additional error types?

accountClient/accountClient.go
willmichael: For point 3 what is the separate authorization pattern that I'm missing, aren't they all using authClient?

lkwdwrd: definitely a conversation worth having. These may well belong in the authClient, though the client service is taking over auth service responsibilities. Things get very messy conceptually as the client service starts to stand up. It may be best to refactor some of these. Also worth noting, the client service using an account client will be getting a token from itself before requesting from another service? Meta...

main.go
lkwdwrd: Worth doing, nice call out.

accountClient/accountClient.go
galxy25: @willmichael no take a look at slides [4-7](https://docs.google.com/presentation/d/1JBrr9XWrT1pocxIe7NgXiOyb5CLYDnxU2wJNrx6gmO8/edit#slide=id.g544c46c84e_3_74)

galxy25: @lkwdwrd  yes it is messy, but that is the job of this client/repo, to abstract the mess behind named methods that take care of the complexity behind the scene and will change as we refactor the backing services.

PR #18, opened 2019-04-16
Fix broken tests and clients to work with non-goish error handling.

galxy25:
- Add pds client method for "health" checking whether the storage service is up.


lkwdwrd:
So -- I regret the way I added the special error type to the methods. It was the wrong way to do it, but I hadn't finished reading GoPL to the point I knew the better way.

We should return the status error as the `error` type instead. It meets the error interface fine. For consuming packages, if they need to use the status error at all, they can use type assertion and response if the returned error is indeed a status error.

I'm sorry I didn't get a PR up, but I figured I had more changes coming... The diff is here: https://github.com/tozny/e3db-clients-go/compare/feature/account-validation?expand=1. I stopped client service work, so it's still partial state.

Either way, I'd like to see us return `error` instead of the custom type. But hey, I'm getting there :)

PR #20, opened 2019-04-22
Feature/pds allowed reads endpoint

willmichael:
companion pr https://github.com/tozny/e3dbSearchService/pull/65

### DIFF
--- a/pdsClient/pdsClient.go
+++ b/pdsClient/pdsClient.go
@@ -149,6 +149,18 @@ func (c *E3dbPDSClient) InternalAllowedReads(ctx context.Context, readerID strin
 	return result, e3dbClients.FlatMapInternalError(*internalError)
 }
 
+// InternalAllowedReadsForAccessPolicy attempts to retrieve the list of allowed readers given an access policy.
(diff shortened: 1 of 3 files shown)


pdsClient/pdsClient.go
```diff
@@ -149,6 +149,18 @@ func (c *E3dbPDSClient) InternalAllowedReads(ctx context.Context, readerID strin
 	return result, e3dbClients.FlatMapInternalError(*internalError)
 }
 
+// InternalAllowedReadsForAccessPolicy attempts to retrieve the list of allowed readers given an access policy.
+func (c *E3dbPDSClient) InternalAllowedReadsForAccessPolicy(ctx context.Context, params InternalAllowedReadersForPolicyRequest) (*InternalAllowedReadersForPolicyResponse, error) {
+	var result *InternalAllowedReadersForPolicyResponse
+	path := c.Host + "/internal/" + PDSServiceBasePath + "/allowed_reads/"
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return result, err
```
lkwdwrd: While not required, I would like us on request failures to provide the status code, and context in the custom error. We should _return_ the error interface, but if the caller needs additional context on the request they can pull pull it out with type assertion.

PR #21, opened 2019-04-24
BatchGetRecords

galxy25:
- Add methods for batch get and proxy batch get of a users records. 
- Integration tests.
- Companion [PR](https://github.com/tozny/pds-service/pull/142)

### DIFF
--- a/pdsClient/api.go
+++ b/pdsClient/api.go
@@ -4,6 +4,8 @@ import (
 	"time"
 )
 
+const BatchReadRecordLimit = 1000
+
 // AddAuthorizedWriterRequest wraps the needed parameters
 // to authorize a client to share records of a specified type on behalf of another client
 type AddAuthorizedWriterRequest struct {
@@ -216,4 +218,13 @@ type InternalGetRecordResponse struct {
 type ListRecordsResult struct {
 	Results   []ListedRecord `json:"results"`
 	LastIndex int            `json:"last_index"`
+}
+
+type BatchGetRecordsRequest struct {
+	RecordIDs   []string `json:"record_ids"`
+	IncludeData bool     `json:"include_data"`
+}
+
+type BatchGetRecordsResult struct {
+	Records []ListedRecord `json:"records"`
 }
--- a/pdsClient/pdsClient.go
+++ b/pdsClient/pdsClient.go
@@ -245,6 +245,33 @@ func (c *E3dbPDSClient) ProxyListRecords(ctx context.Context, authToken string, 
 	return result, e3dbClients.FlatMapInternalError(internalError)
 }
 
+// BatchGetRecords makes a call to batch get the records in params,
+// returning the records (if they exist) and error (if any).
+func (c *E3dbPDSClient) BatchGetRecords(ctx context.Context, params BatchGetRecordsRequest) (*BatchGetRecordsResult, error) {
+	var result *BatchGetRecordsResult
+	path := c.Host + "/" + PDSServiceBasePath + "/records"
+	request, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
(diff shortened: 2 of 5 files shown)


pdsClient/api.go
```diff
@@ -217,3 +217,12 @@ type ListRecordsResult struct {
 	Results   []ListedRecord `json:"results"`
 	LastIndex int            `json:"last_index"`
 }
+
+type BatchGetRecordsRequest struct {
+	RecordIDs   []string `json:"record_ids"`
+	IncludeData bool     `json:"include_data"`
+}
+
+type BatchGetRecordsResult struct {
```
willmichael: What happens a single record fails?

pdsClient/pdsClient.go
```diff
@@ -245,6 +245,33 @@ func (c *E3dbPDSClient) ProxyListRecords(ctx context.Context, authToken string,
 	return result, e3dbClients.FlatMapInternalError(internalError)
 }
 
+// BatchGetRecords makes a call to batch get the records in params,
```
willmichael: Not a proxied user call and not an `internal` endpoint, so who's the consumer of this function? Or is it just a testing helper?

pdsClient/api.go
galxy25: A record "failure" means either the record doesn't exist or the reader can't read it so it won't come back in the results, its the responsibility of the caller to determine whether that was the case, and which one didn't get returned(by doing some set operations on the requested id's versus the returned id's), and then call non batch get for the record(s) of interest to determine whether 404 or 401'd

pdsClient/pdsClient.go
galxy25: It tests the imperfect Scala code I wrote to make, follows the development of the feature starting with implementing the batch pds endpoint and then writing this to test it faster before writing the scala integration test. And now, if we ever have need for a client to get a batch of records in go, the code is ready and tested.

pdsClient/pdsClient.go
```diff
@@ -245,6 +245,33 @@ func (c *E3dbPDSClient) ProxyListRecords(ctx context.Context, authToken string,
...
+// returning the records (if they exist) and error (if any).
+func (c *E3dbPDSClient) BatchGetRecords(ctx context.Context, params BatchGetRecordsRequest) (*BatchGetRecordsResult, error) {
+	var result *BatchGetRecordsResult
+	path := c.Host + "/" + PDSServiceBasePath + "/records/batch"
```
lkwdwrd: would this endpoint better live at `GET /records` instead of an additional sub-path that is not win URL namespace competition with `/records/<record-id>`? Also seems like this can be the genesis of fully deprecating v1/serach and moving it to a list service, available at `GET /records`?

Is the request payload in the body or URL params?

pdsClient/pdsClient_test.go
```diff
@@ -73,6 +73,96 @@ func MakeClientWriterForRecordType(pdsUser E3dbPDSClient, clientID string, recor
 
 var validPDSEncryptedAccessKey, _ = MakeClientWriterForRecordType(validPDSUser, validPDSUserID, defaultPDSUserRecordType)
 
+func TestBatchGetRecordsReturnsBatchofUserRecords(t *testing.T) {
```
lkwdwrd: Seeing happy path testing, but no sad path testing. Would like to see a couple sad path tests to ensure failures happen as we expect.

pdsClient/pdsClient.go
galxy25: Good call on `GET /records`. 
Yeah, I think between this and v2 search we could deprecate v1 search and not lose any features for the customer.
Request payload is in the body.

pdsClient/pdsClient_test.go
galxy25: added some more unhappy path testing.

PR #22, opened 2019-04-29
Add method and integration test for internal allowed_reads search endpoint.

galxy25:
- Companion [PR](https://github.com/tozny/pds-service/pull/146)
- Promote InternalModifiedRecords to more fleshed out internal search method.

### DIFF
--- a/pdsClient/api.go
+++ b/pdsClient/api.go
@@ -64,21 +64,24 @@ type AllowedRead struct {
 	ContentType string `json:"content_type"`
 }
 
-// InternalModifiedSearchRequest represents a valid request to internal records search for getting modified records
-type InternalModifiedSearchRequest struct {
-	NextToken int                   `json:"next_token"`
-	Range     InternalModifiedRange `json:"range"`
+// InternalSearchRequest represents a valid request to the internal records search endpoint
+type InternalSearchRequest struct {
+	NextToken    int                    `json:"next_token,omitempty"`
+	Range        *InternalModifiedRange `json:"range,omitempty"`
+	WriterIDs    []string               `json:"writer_ids,omitempty"`    //If not empty, limit results to records written by the given set of IDs.
+	UserIDs      []string               `json:"user_ids,omitempty"`      //If not null, limit results to records about given set of IDs.
+	ContentTypes []string               `json:"content_types,omitempty"` //If not empty, limit results to records of the given types.
 }
 
-// InternalModifiedSearchResponse represents a response from calling the internal records search endpoint
-type InternalModifiedSearchResponse struct {
+// InternalSearchResponse represents a response from calling the internal records search endpoint
+type InternalSearchResponse struct {
 	NextToken int      `json:"next_token"`
 	Records   []Record `json:"records"`
 }
 
 // InternalSearchAllowedReadsRequest represents a valid request to internal allowed reads search
 // for getting a list of allowed reads that match the request parameters.
-type InternalSearchAllowedReadsRequest = InternalModifiedSearchRequest
+type InternalSearchAllowedReadsRequest = InternalSearchRequest
 
 // InternalSearchAllowedReadsResponse represents a response from calling the internal allowed reads search endpoint
 type InternalSearchAllowedReadsResponse struct {
--- a/pdsClient/api.go
+++ b/pdsClient/api.go
@@ -76,6 +76,16 @@ type InternalModifiedSearchResponse struct {
 	Records   []Record `json:"records"`
 }
 
+// InternalSearchAllowedReadsRequest represents a valid request to internal allowed reads search
+// for getting a list of allowed reads that match the request parameters.
+type InternalSearchAllowedReadsRequest = InternalModifiedSearchRequest
+
+// InternalSearchAllowedReadsResponse represents a response from calling the internal allowed reads search endpoint
+type InternalSearchAllowedReadsResponse struct {
+	NextToken    int           `json:"next_token"`
+	AllowedReads []AllowedRead `json:"allowed_reads"`
+}
+
 // InternalModifiedRange represents the range parameter for the internal modified search request
 type InternalModifiedRange struct {
 	After  time.Time `json:"modified_after"`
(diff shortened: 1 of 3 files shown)


pdsClient/pdsClient.go
```diff
@@ -173,9 +173,10 @@ func (c *E3dbPDSClient) InternalRegisterClient(ctx context.Context, params Regis
 	return result, e3dbClients.FlatMapInternalError(*internalError)
 }
 
-// InternalModifiedSearch uses and internal endpoint to search for modified records within a time range
-func (c *E3dbPDSClient) InternalModifiedSearch(ctx context.Context, params InternalModifiedSearchRequest) (*InternalModifiedSearchResponse, error) {
-	var result *InternalModifiedSearchResponse
+// InternalSearch returns records macthing the provided params,
```
willmichael: *matching

pdsClient/api.go
```diff
@@ -64,18 +64,31 @@ type AllowedRead struct {
...
 }
 
-// InternalModifiedSearchRequest represents a valid request to internal records search for getting modified records
-type InternalModifiedSearchRequest struct {
-	NextToken int                   `json:"next_token"`
-	Range     InternalModifiedRange `json:"range"`
+// InternalSearchRequest represents a valid request to the internal records search endpoint
+type InternalSearchRequest struct {
+	NextToken    int                    `json:"next_token,omitempty"`
+	Range        *InternalModifiedRange `json:"range,omitempty"`
```
willmichael: Range can also be empty right? If I wanted to just search on writerIDs/UserID/ContentType would it also work?

pdsClient/api.go
```diff
@@ -64,18 +64,31 @@ type AllowedRead struct {
...
 	Records   []Record `json:"records"`
 }
 
+// InternalSearchAllowedReadsRequest represents a valid request to internal allowed reads search
+// for getting a list of allowed reads that match the request parameters.
+type InternalSearchAllowedReadsRequest = InternalSearchRequest
+
+// InternalSearchAllowedReadsResponse represents a response from calling the internal allowed reads search endpoint
+type InternalSearchAllowedReadsResponse struct {
+	NextToken    int           `json:"next_token"`
```
willmichael: Doesnt this endpoint also need a range? We're looking for modified allowed reads within a certain time range?

pdsClient/pdsClient.go
```diff
@@ -185,6 +186,19 @@ func (c *E3dbPDSClient) InternalModifiedSearch(ctx context.Context, params Inter
...
+// provided params, returning a paginated list of allowed reads and error (if any).
+func (c *E3dbPDSClient) InternalSearchAllowedReads(ctx context.Context, params InternalSearchAllowedReadsRequest) (*InternalSearchAllowedReadsResponse, error) {
+	var result *InternalSearchAllowedReadsResponse
+	path := c.Host + "/internal/" + PDSServiceBasePath + "/allowed_reads"
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return result, err
+	}
+	internalError := e3dbClients.MakeE3DBServiceCall(c.E3dbAuthClient, ctx, request, &result)
+	return result, e3dbClients.FlatMapInternalError(*internalError)
```
lkwdwrd: As this `FlatMap` function continues to show up more often, the more I want to update this code base to remove this custom error type from direct returns...

pdsClient/pdsClient_test.go
```diff
@@ -731,3 +731,88 @@ func TestServiceHealthCheckReturnsSuccessIfPDSIsRunning(t *testing.T) {
...
+	recordToWrite := WriteRecordRequest{
+		Data: data,
+		Metadata: Meta{
+			Type:     recordType,
+			WriterID: validPDSUserID,
+			UserID:   validPDSUserID,
+			Plain:    map[string]string{"key": "value"},
+		},
+	}
+	ctx := context.TODO()
```
lkwdwrd: Shouldn't this be `context.Background()` from a semantic standpoint? As I currently understand context, both are noop contexts, but background is "This is noop intentionally" while TODO is "This code is being refactored to use context, TODO is make it work with it correctly".

pdsClient/api.go
galxy25: Correct

pdsClient/pdsClient.go
galxy25: 🦅 👁

pdsClient/pdsClient.go
galxy25: Agree, I think we can keep the usefulness of the custom error by having the `MakeE3DBServcieCall` function return type error(returning nil if no error, and the RequestError struct otherwise)  and having callers of these methods use runtime type assertion to cast it to a ResponseError as needed/desired.

pdsClient/pdsClient_test.go
galxy25: Good point, updated.

galxy25:
@lkwdwrd 
> Seems pretty straight forward. +1 on Michaels comments.
> 
> In general I'm struggling a bit with what I see in our tests. I still see a lot of happy-path only testing. The struggle for this one in particular is that we are not testing the API specifically, we are trying to test the SDK (Cleint) that makes requests to the API. Where is the right place to check error handling? Not entirely sure. Maybe happy path (does this call the endpoint and encode/decode the JSON correctly) is enough. I'll be pondering this one.

In general, non happy/negative tests are harder to write and outside of security/access checks have low return on investment, which is why I think for even new packages written by someone as intent on non-happy path testing as [yourself](https://github.com/tozny/storage-service/blob/initial-service/service/main_test.go) they tend to fall by the way side. 

For this package specifically, happy pass testing is the best/most useful thing we can do - these client exists to call endpoints, and the integration tests are against running versions of those services; if we wanted to test failure case we would have to create mock services to call, which would be useful unit tests to make sure we handle errors gracefully -imo the benefit gained from such unit tests is not worth the effort. For what its work, there is non happy path testing of this endpoint in the codebase the endpoint is implemented in, testing the non happy paths I considered to be important enough(making sure external clients can't access it, making sure it doesn't return data when it shouldn't)

PR #23, opened 2019-04-29
Add methods for account internal get stripe id

efabens:
- Add method for new account service get stripe id internal endpoint
- Adds tests for those methods
- Refactors test file to not have duplicate code
- [Companion PR](https://github.com/tozny/account-service/pull/58)

### DIFF
--- a/accountClient/accountClient.go
+++ b/accountClient/accountClient.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"github.com/tozny/e3db-clients-go"
-	"github.com/tozny/e3db-clients-go/authClient"
 	"net/http"
+
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


accountClient/accountClient.go
```diff
@@ -43,6 +43,18 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
 	return result, err
 }
 
+// InternalGetStripeID attempts to get the stripe id for the specified account id
+func (c *E3dbAccountClient) InternalGetStripeID(ctx context.Context, accountID string) (*InternalGetStripeIDResponse, error) {
```
willmichael: https://golang.org/doc/effective_go.html#Getters Not necessarily suggesting a change here as I also prefer `Get` in many places, but maybe worth noting the idiomatic go way of naming getters.

galxy25: @willmichael yes but in this case these are http clients so they're (very loosely) following the idiom of RESTful design

efabens:
Added a new field and rebasing was a disaster so having another quick look over wouldn't hurt

PR #26, opened 2019-05-07
Feature/account validation

willmichael:
Cleanup work done to remove typed error handling done in conjunction with this PR https://github.com/tozny/client-service/pull/2/files


accountClient/accountClient.go
```diff
@@ -20,19 +22,19 @@ type E3dbAccountClient struct {
 }
 
 // CreateAccount attempts to create an e3db account using the provided params, returning created account and error (if any).
-func (c *E3dbAccountClient) CreateAccount(ctx context.Context, params CreateAccountRequest) (*CreateAccountResponse, *e3dbClients.RequestError) {
+func (c *E3dbAccountClient) CreateAccount(ctx context.Context, params CreateAccountRequest) (*CreateAccountResponse, error) {
```
galxy25: I don't think you can make this change without breaking any code that tries to do things like `https://github.com/tozny/metrics-service/commit/a16d7ee99700c75dd4a66bf5daa95267a01766f6#diff-c47669012d7664f9cbee69c67bf8bedfR90` so make sure you test this change out for those cases

accountClient/api.go
```diff
@@ -66,3 +68,27 @@ type CreateAccountResponse struct {
...
+// ValidateTokenResponse represents the result of calling the account service's auth/validate endpoint.
+type ValidateTokenResponse struct {
+	AccountID string `json:"account_id"` //The account ID associated with this token
+	Valid     bool   `json:"valid"`      //Whether the token was valid
+}
+
+// RegTokenInfo is the return from the token endpoint on a valid request
+type RegTokenInfo struct {
+	Token       string
+	AccountID   uuid.UUID `json:"account_id"`
```
galxy25: why not a string like the other AccountID at line 79?

lkwdwrd:
I'm very happy to see this come up. I've regretted directly exposing this custom error type since I learned more about error handling in golang. I agree it may not be 100% compatible with a code base that is expecting the custom struct as an error return. The consuming code can be very quickly updated to use type assertion to pull out the correct type if it does need access to the custom type. Tagging versions correctly should help us avoid major conflicts I would think.

@galxy25 - happy to discuss further, I though you and I were on the same page in terms stopping directly exposing the custom error object, but perhaps I am wrong?

galxy25:
@lkwdwrd talked to 
@willmichael in person, I'm on the same page now and onboard(again)

accountClient/api.go
willmichael: @lkwdwrd Preference here?

lkwdwrd: No preference -- let's stay consistent

PR #28, opened 2019-05-17
MetricClient to call the new endpoints for the Metrics-Service

willmichael:
Code to call the new endpoints for the metrics service

Change from IncludeLogs to ExcludeLogs that will be updated in the spec because for booleans golang defaults to "false" if not provided. There's no way to easily differentiate if the user set "false" or they didn't provide a value that I'm aware of. You can do some tricks with a bool ptr and omitempty, but it leads to an unclean solution (unless someone else has the answer to this)


go.sum
```diff
@@ -1,73 +1,172 @@
...
+github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+github.com/tozny/e3db-clients-go v0.0.9/go.mod h1:WZ/FZ8gl5fSafMA5MCt37fMg0Nh5YBDlxErLvZNKYe8=
+github.com/tozny/e3db-clients-go v0.0.18-0.20190502231832-3477741e5116/go.mod h1:3IhfF+eO1GvhN6ugx8BzqXVQIYpvFcYgL3I01mNSv7E=
+github.com/tozny/e3db-clients-go v0.0.19-0.20190507195351-473922bdcefc/go.mod h1:Pn8q5elypmNENO9ARBulYeicvN2vuuOaupl+O+DXlp8=
+github.com/tozny/e3db-clients-go v0.0.19/go.mod h1:4/0vmSE45SXY5Qh6X2RLCcuaFUQ4IVeoMl5wJXRNBfE=
 github.com/tozny/e3db-go/v2 v2.1.1 h1:esYqwmaxUpvIS0R7lcvifyfERSx/fSWs3fYu+5GQHR4=
 github.com/tozny/e3db-go/v2 v2.1.1/go.mod h1:fm4qLvM+lVyILrB+GEC/1pmwHXlzPFpY7+fB9RB6xio=
+github.com/tozny/utils-go v0.0.4/go.mod h1:eEO5khjSLcWQwTtwuu8ANsERmyh8szbukGSmgCAVjlU=
```
galxy25: Hmm that looks quite odd, can you run `go mod tidy` and re-push this file when you get a chance

PR #30, opened 2019-05-24
namespace change to append /requests for current Metrics endpoints

willmichael:
namespace change to match this PR https://github.com/tozny/api-specs/pull/83


willmichael:
Closing with 1 approval because it's a relatively small change that shouldn't impact other functionality and I need to a relevant tag in another branch

PR #33, opened 2019-06-07
Initial billing service implementation

efabens:
Tests will get added in a later implementation, this sprint

### DIFF
--- a/billingClient/billingClient.go
+++ b/billingClient/billingClient.go
@@ -1,0 +1,41 @@
+package billingclient
+
+import (
+	"context"
+
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


billingClient/billingClient.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
+)
+
+const (
+	BillingServiceBasePath = "v1/billing"
+	InternalCreateCustomer = "/internal/create_customer"
```
galxy25: Internal paths should follow the current pattern of `internal/apiVersion/serviceName/path`.

galxy25: Unless you forsee this changing a lot/being used by consumers of this package I don't thank this needs to be a constant or public.

efabens: 👍 will update

PR #34, opened 2019-06-07
Account Registration Endpoints

willmichael:
Adds registration and creation of registration tokens to the accountClient. However account service comes with its own Auth token, and it serves up extra data information in the headers of the response which we abstract out with e3db-clients-go leading to near duplicate functions. Suggestions on how to handle this welcome, focus was more on getting client-service working than making this client super clean so there might be a really simple way I've overlooked

companion PR https://github.com/tozny/client-service/pull/7

### DIFF
--- a/accountClient/accountClient.go
+++ b/accountClient/accountClient.go
@@ -46,6 +46,35 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
 	return result, err
 }
 
+func (c *E3dbAccountClient) RegisterClient(ctx context.Context, params ClientRegistrationRequest) (*ClientRegistrationResponse, error) {
+	var result *ClientRegistrationResponse
+	path := c.Host + "/" + AccountServiceBasePath + "/e3db/clients/register"
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	resp, err := e3dbClients.ReturnE3dbServiceCall(c.E3dbAuthClient, ctx, request, &result)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	backupClient := resp.Header.Get("X-Backup-Client")
+	result.RootClientID = backupClient
+	return result, err
+}
+
+func (c *E3dbAccountClient) CreateRegistrationToken(ctx context.Context, accountServiceJWT string, params CreateRegTokenRequest) (*CreateRegTokenResponse, error) {
+	var result *CreateRegTokenResponse
+	path := c.Host + "/" + AccountServiceBasePath + "/tokens"
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	// TODO: Not actually a `proxied user call` but account service serves it's own auth...
+	// Consider a renaming of MakeProxiedUserCall
+	err = e3dbClients.MakeProxiedUserCall(ctx, accountServiceJWT, request, &result)
+	return result, err
+}
+
 // RegistrationToken validates a registration token with the account service and fetches its permissions
 func (c *E3dbAccountClient) RegistrationToken(ctx context.Context, token string) (*RegTokenInfo, error) {
 	result := RegTokenInfo{
--- a/accountClient/accountClient.go
+++ b/accountClient/accountClient.go
@@ -46,6 +46,8 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
 	return result, err
 }
 
+// RegisterClient registers a client with using the account service which proxies to the client service,
+// this method is intended for TESTING the functionality of the integrated client service. Not intended for future use.
 func (c *E3dbAccountClient) RegisterClient(ctx context.Context, params ClientRegistrationRequest) (*ClientRegistrationResponse, error) {
 	var result *ClientRegistrationResponse
 	path := c.Host + "/" + AccountServiceBasePath + "/e3db/clients/register"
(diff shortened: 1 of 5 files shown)


accountClient/accountClient.go
```diff
@@ -46,6 +46,35 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
...
+	resp, err := e3dbClients.ReturnE3dbServiceCall(c.E3dbAuthClient, ctx, request, &result)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	backupClient := resp.Header.Get("X-Backup-Client")
+	result.RootClientID = backupClient
+	return result, err
+}
+
+func (c *E3dbAccountClient) CreateRegistrationToken(ctx context.Context, accountServiceJWT string, params CreateRegTokenRequest) (*CreateRegTokenResponse, error) {
```
galxy25: either accountServiceJWT should be part of the struct for an `E3dbAccountClient` or it should be part of the params struct.

accountClient/accountClient.go
```diff
@@ -46,6 +46,35 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
...
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	resp, err := e3dbClients.ReturnE3dbServiceCall(c.E3dbAuthClient, ctx, request, &result)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	backupClient := resp.Header.Get("X-Backup-Client")
+	result.RootClientID = backupClient
```
galxy25: Looks to me like we can get rid of `ReturnE3dbServiceCall` if we pass `c.E3dbAuthClient.AuthHTTPClient(ctx)` to `MakeRawServiceCall` and have register client return the backup client for a newly registered client as part of the response object(which is I change we should make regardless of making this easier to refactor, @lkwdwrd any context on why that value is in a header instead of the response body?)

accountClient/api.go
```diff
@@ -81,6 +101,24 @@ type ValidateTokenResponse struct {
...
+type CreateRegTokenResponse struct {
+	Token       string           `json:"token"`
+	Permissions TokenPermissions `json:"permissions"`
+}
+
+// TokenPermissions permissions associated with a registration token.
+// called ClientPermissions in the account service spec
+type TokenPermissions struct {
+	Enabled bool `json:"enabled"`  // Flag a newly created client as enabled even if the default behavior is creating disabled clients
+	OneTime bool `json:"one-time"` // Automatically delete the token after it's been used to register a client
```
galxy25: snake case for JSON names

accountClient/accountClient.go
lkwdwrd: not really. I think the idea is this was response meta rather than response data, hence the header. The response could be handled one way, and the backup ID could could be fetched separately if desired by the SDKs. At this point I don't think we should change it, since this would break our current SDKs. Going forward, I think this will shift to the returned client info, which is part of the response body. Until that time, let's not make a breaking change to the API.

accountClient/api.go
lkwdwrd: I'll have to check, but I'm pretty sure I changed this to snake case in the account service refactor. I agree snake case is correct. It was in the API before, but unused, now it is used for the registration flow. Changing it is not any kind of a breaking change.

accountClient/accountClient.go
willmichael: specifically for testing, AFAIK the token only comes from a register request to account service?

accountClient/accountClient.go
```diff
@@ -46,6 +46,35 @@ func (c *E3dbAccountClient) InternalGetClientAccount(ctx context.Context, client
 	return result, err
 }
 
+func (c *E3dbAccountClient) RegisterClient(ctx context.Context, params ClientRegistrationRequest) (*ClientRegistrationResponse, error) {
+	var result *ClientRegistrationResponse
+	path := c.Host + "/" + AccountServiceBasePath + "/e3db/clients/register"
```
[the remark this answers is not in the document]
willmichael: We have a client-service client to query the client service, purely written for testing purposes

accountClient/accountClient.go
galxy25: Right, I'm not arguing for making a breaking change, I'm suggesting we make an additive change(return the backup client as part of the new client service response), let the sdk continue to use the header value, but don't write any new code(including wherever we are calling this method) that expects this value to be present in a header.

accountClient/api.go
willmichael: Taken from the current spec, but since it's been changed I'll update it

accountClient/accountClient.go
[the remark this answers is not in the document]
lkwdwrd: Eventually the account service version of this endpoint will get removed -- it will probably be a little while before everything is relying on the new endpoint, but in general we should not write any new code relying on the deprecated endpoints. If we need this for compatibility, then it's fine, but if this is new testing functionality then we should either use the client service client, or this should hit the client service endpoint.

I may not be understanding what this is, too, so feel free to correct me if need be :)

accountClient/api.go
lkwdwrd: I do have it updated the account service to use snake-case. The original code was camel case, but agree, underscores for JSON names is better all around.

accountClient/accountClient.go
lkwdwrd: I guess I worry about breaking things like the Java and other strongly typed SDKs by adding values to the return... The JSON parsing is not as fault-tolerant as in go. We are using a different JSON library in the SDK than in the Scala code, so maybe it works better, but I know in the scala parser, if the JSON doensn't fit into the box perfectly it starts throwing errors...

@efabens - do you happen to know off the top of your head if the JSON parsing in the Java SDK would properly deal with an extra field in the JSON return in a registration response? (if not off the top of your head, don't worry about it).

(it will probably be OK. We did add an Access Key ID to record returns and nothing got too angry)

accountClient/accountClient.go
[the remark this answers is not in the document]
willmichael: Just an added layer for me to make sure the registration implementation was working. We had the from account service to client service so I wanted to make sure callers of the account service would get back expected results. For the future client service should be used. I updated this with a comment that it should be used for testing, and I can remove this method now or later if you like

accountClient/api.go
willmichael: updated 👍

accountClient/accountClient.go
willmichael: For this PR I will do a no-op, and make a note to refactor once account service is updated

PR #35, opened 2019-06-07
Client-service client

willmichael:
Adds new client for the client-service, to call the four new endpoints

companion PR https://github.com/tozny/client-service/pull/7

### DIFF
--- a/clientServiceClient/api.go
+++ b/clientServiceClient/api.go
@@ -14,6 +14,31 @@ type AdminListResponse struct {
 	NextToken int64    `json:"next_token"`
 }
 
+type AdminGetResponse struct {
+	Client
+}
+
+type ClientGetResponse struct {
+	Client
+}
+
+// ClientRegisterRequest captures the information sent to create a client.
+type ClientRegisterRequest struct {
+	RegistrationToken string `json:"token"`
+	Client            struct {
+		Name       string            `json:"name"`
+		Type       string            `json:"type"`
+		PublicKey  map[string]string `json:"public_key"`
+		SigningKey map[string]string `json:"signing_key,omitemtpy"`
+	} `json:"client"`
+}
+
+// ClientRegisterResponse sends back the JSON information of a client.
+type ClientRegisterResponse struct {
+	Client
+	APISecret string `json:"api_secret"`
+}
+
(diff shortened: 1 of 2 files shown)


clientServiceClient/api.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type ClientGetResponse struct {
+	Client
+}
+
+// ClientRegisterRequest captures the information sent to create a client.
+type ClientRegisterRequest struct {
+	RegistrationToken string `json:"token"`
+	Client            struct {
```
galxy25: why this inline struct versus using the client type on line 43 below?

clientServiceClient/api.go
```diff
@@ -0,0 +1,52 @@
...
+		Name       string            `json:"name"`
+		Type       string            `json:"type"`
+		PublicKey  map[string]string `json:"public_key"`
+		SigningKey map[string]string `json:"signing_key,omitemtpy"`
+	} `json:"client"`
+}
+
+// ClientRegisterResponse sends back the JSON information of a client.
```
galxy25: nit: JSON is a redundant/junk food word in this comment. Suggest replacing it with something more information dense like "sends back the client information for a newly registered client"

clientServiceClient/api.go
willmichael: This inline struct doesn't accept these as fields

	ClientID    uuid.UUID         `json:"client_id"`
	APIKeyID    string            `json:"api_key_id"`
	Name        string            `json:"name"`
	Enabled     bool              `json:"enabled"`
	Meta        map[string]string `json:"meta,omitempty"`

galxy25: Right, but why do we need a separate struct versus having the endpoint that processes this request only extract the fields of interest?

willmichael: To provide the clarity of a specific type, what are we gaining by using the separate struct with additional information except for saving lines? I agree this inline struct is a little ugly, but the fields the user provides will actually be used.

galxy25: As long as we don't need this representation of a client elsewhere, my answer to your question is: `nothing`.

PR #37, opened 2019-06-12
client service public endpoint

willmichael:
Companion PR here: https://github.com/tozny/client-service/pull/9. Adds client functions to call new endpoints

### DIFF
--- a/clientServiceClient/api.go
+++ b/clientServiceClient/api.go
@@ -28,13 +28,18 @@ type ClientGetResponse struct {
 	Client
 }
 
+// ClientGetPublicResponse is the client information from the endpoint /<client_id>/public
+type ClientGetPublicResponse struct {
+	PublicClient
+}
+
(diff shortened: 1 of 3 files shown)


clientServiceClient/api.go
```diff
@@ -28,15 +28,23 @@ type ClientGetResponse struct {
...
 type ClientRegisterRequest struct {
-	RegistrationToken string `json:"token"`
-	Client            struct {
-		Name       string            `json:"name"`
-		Type       string            `json:"type"`
-		PublicKey  map[string]string `json:"public_key"`
-		SigningKey map[string]string `json:"signing_key,omitemtpy"`
-	} `json:"client"`
+	RegistrationToken string             `json:"token"`
+	Client            ClientRegisterInfo `json:"client"`
```
willmichael: @galxy25 looks like you were right about inline structs, initializing them is a pain 😬

PR #38, opened 2019-06-12
initial internal patch implemenation

willmichael:
Update Client service client to have internal patch endpoint: 
companion PR: https://github.com/tozny/client-service/pull/10

### DIFF
--- a/clientServiceClient/clientServiceClient.go
+++ b/clientServiceClient/clientServiceClient.go
@@ -42,6 +42,17 @@ func (c *ClientServiceClient) AdminGet(ctx context.Context, clientID string) (*A
 	return result, err
 }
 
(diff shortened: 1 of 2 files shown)


clientServiceClient/clientServiceClient.go
```diff
@@ -78,6 +78,17 @@ func (c *ClientServiceClient) GetPublicClient(ctx context.Context, clientID stri
 	return result, err
 }
 
+// InternalPatchBackup gets a client's public client information for clientID.
```
galxy25: Why is this namespaced as internal if this is public information?

willmichael: bad copy pasta comment 😬

PR #39, opened 2019-06-12
Initial batch public endpoints

willmichael:
Adds client calls for the internal batchpublic endpoint to get a client's public information

Companion PR: https://github.com/tozny/client-service/pull/11

### DIFF
--- a/clientServiceClient/api.go
+++ b/clientServiceClient/api.go
@@ -33,6 +33,16 @@ type ClientGetPublicResponse struct {
 	PublicClient
 }
 
+// ClientBatchPublicRequest is a list of clientIDs to get public information for.
+type ClientBatchPublicRequest struct {
+	ClientIDs []string `json:"client_ids"`
+}
+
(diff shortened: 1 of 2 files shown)


clientServiceClient/api.go
```diff
@@ -33,6 +33,16 @@ type ClientGetPublicResponse struct {
 	PublicClient
 }
 
+// ClientBatchPublicRequest is a list of clientIDs to get public information for.
+type ClientBatchPublicRequest struct {
```
galxy25: suggest: `BatchClientInfoRequest` or `BatchClientPublicRequest` , I'm not a fan of naming things in a way that leads to more questions(its clear to you today with public means but it will it be so obvious in the future for you and others?(not withstanding the helpful doc comment))

willmichael: 👍 went for public to differentiate a standard Info request from a public request, but maybe even further: `BatchClientPublicInfoRequest` is the way to go.

PR #41, opened 2019-06-14
Add initial identity service client with service and health check methods.

galxy25:


identityClient/identityClient_test.go
```diff
@@ -0,0 +1,32 @@
...
+	e3dbClients "github.com/tozny/e3db-clients-go"
+)
+
+var (
+	e3dbIdentityHost          = os.Getenv("E3DB_IDENTITY_SERVICE_HOST")
+	ValidIdentityClientConfig = e3dbClients.ClientConfig{
+		Host: e3dbIdentityHost,
+	}
+	identityServiceClient = New(ValidIdentityClientConfig)
+	testContext           = context.TODO()
```
lkwdwrd: I'm still far from an expert on context -- but as I understand it this is something that should be 100% request scoped. Having this created globally is probably fine for testing, but may confuse newer Toznian Gophers in terms of how context works, which is already confusing. It means more copy pasta, but I'd vote for creating a `context.Backgrount()` inline for each as a demonstration, rather than creating a global context object.

galxy25: Right, context should be some discrete scope, like for a server each request gets a separate context. In this case I'd argue the test file is a single context, with go being able to run each test function in parallel, and while we're not cancelling any of the contexts, its likely that if the context in one test fails we want the tests to abort/or if not one would explicitly set up a different context for that case. New gopher's are going to struggle with context no matter what I think, based on my own continuing struggles 😭 with it.

PR #48, opened 2019-07-08
Enable management of realms using e3db-clients-go

galxy25:
- Add identity service realm management methods and tests.
- Update client config to contain signing and encryption keypairs. 
- Update authNInfo response to include the client's type.
- Update readme variables to current url's for platform services.
- Consolidate client-service and e3db-clients-go RegisterClient test helper.


identityClient/identityClient.go
```diff
@@ -16,10 +18,24 @@ type E3dbIdentityClient struct {
...
+// CreateRealm creates a realm using the specified parameters,
+// returning the created realm (including it's associated sovereign) and error (if any).
+func (c *E3dbIdentityClient) CreateRealm(ctx context.Context, params CreateRealmRequest) (*CreateRealmResponse, error) {
+	var createRealmResponse *CreateRealmResponse
+	path := c.Host + identityServiceBasePath + "/" + realmResourceName
+	request, err := e3dbClients.CreateRequest("POST", path, params)
+	if err != nil {
+		return createRealmResponse, err
+	}
+	// TODO: signature based request
```
galxy25: (before the pr/ sprint story is clone)

identityClient/api.go
```diff
@@ -0,0 +1,25 @@
+package identityClient
+
+// Sovereign represents the top level user of a realm
+// (i.e. a user who is able to log into the realm,
+// create/update/delete/view the realm properties and it's identities.)
+type Sovereign struct {
+	Id       int64  `json:"id"`       // Service defined unique identifier for the sovereign.
+	Name     string `json:"name"`     // User defined sovereign identifier.
+	Password string `json:"password"` // The sovereign's realm administration console login password.
```
lkwdwrd: Does the Sovereign need a password if we can authenticate a signed request in the Keycloak instance?

galxy25: Good question. For now, yes, because the sovereign needs to be able to log into their realm/keycloak, which is what we're doing for mvp(@eddier) . Post launch though I want to write a keycloak extension/ authentication  SPI that will allow sovereigns to log into their realms/keycloak using their tozny account. Currently the sovereign is solely for interacting with the realm, while programatic access and control of Tozny Identity Management Service resources such as realms, sovereigns, consuming identities(consuming app users) is done with a (yet to be built) tozny sdk(s) identity client (which will use their existing tozny API credentials/keys) by talking to the identity service api which acts as the control plane/abstraction over the keycloak API.

identityClient/identityClient.go
```diff
@@ -16,10 +19,61 @@ type E3dbIdentityClient struct {
 	authClient *http.Client
 }
 
+// ListRealms lists the realms belonging to the requester returning the realms and error (if any).
```
willmichael: does the endpoint provide authZ based on identifying information in the signature?

main.go
```diff
@@ -8,6 +8,31 @@ import (
...
+const (
+	// https://security.stackexchange.com/questions/50878/ecdsa-vs-ecdh-vs-ed25519-vs-curve25519
+	DefaultEncryptionKeyCurve = "curve25519"
+	DefaultSigningKeyCurve    = "ed25519"
+)
+
+// Key wraps material generated using an algorithm/curve for use in cryptographic operations
+type Key struct {
+	Material string
+	Curve    string
```
willmichael: Curve -> Type

main.go
```diff
@@ -16,7 +41,9 @@ type ClientConfig struct {
 	APISecret string // User/Client secret to use when communicating with the e3db API
 	// Hostname for the soon to be deprecated (v1) e3db bearer auth service API.
 	// Once request signing is the primary mode of authenticating e3db requests this can be removed.
-	AuthNHost string
+	AuthNHost      string
+	SigningKeys    SigningKeys    // AsymmetricEncryptionKeypair used for signing and authenticating requests
+	EncryptionKeys EncryptionKeys // AsymmetricEncryptionKeypair used for encrypting and decrypting data
```
willmichael: Does a request client need encryption keys?

identityClient/identityClient.go
galxy25: yes, its protected by cyclops to enforce only queen client tsv1 authenticated requests arrive at this endpoint.

main.go
galxy25: No, encryption keys are used for client side operations such as encrypting the record before writing(not currently implemented) and writing access keys for other users

PR #50, opened 2019-07-16
Account client Info endpoint

willmichael:
Adds client call to hit the internal endpoint /info for account service/metrics service


willmichael:
> I would prefer the /internal-accessing methods be namespaced as "InternalDoX" like the "InternalGetClientAccount" method, but we already don't have consistency on that, so it's definitely not blocking.
@emanb29 
Good to move towards consistency, thanks for pointing it out

PR #52, opened 2019-07-23
Add type safety to keys and crypto methods to support tozstore-like operations

willmichael:
Defines some shared types used across services like the additional tozny client contextual information and public encryption keys.

Ports over crypto methods from our SDK so we can use proper encryption in Integration tests

Adds storage client to for use in integration tests against storagev2/notes service.

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -54,10 +54,10 @@ type EncryptionKeys AsymmetricKeypair
 // or externally (client calls).
 type PublicEncryptionKeys Keys
 
-// PublicSignatureKeys are a set of key, NIST or Sodium, used within Tozny to sign requests
+// PublicSigningKeys are a set of key, NIST or Sodium, used within Tozny to sign requests
 // It's the responsibility of the user to assign proper key types to raw keys obtained internally (db)
 // or externally (client calls).
-type PublicSignatureKeys Keys
+type PublicSigningKeys Keys
 
 // GenerateSigningKeys generates a `base64.RawURLEncoding` private and public key
 // for signing requests and data on behalf of Tozny clients,
--- a/crypto.go
+++ b/crypto.go
@@ -14,9 +14,9 @@ const (
 	DefaultSigningKeyType    = "ed25519"
 )
 
-// ToznyAuthenticatedContext represents the contextual information provided by cyclops to downstream services
+// ToznyAuthenticatedClientContext represents the contextual information provided by cyclops to downstream services
 // when a user is successfully authenticated.
-type ToznyAuthenticatedContext struct {
+type ToznyAuthenticatedClientContext struct {
 	ClientID       uuid.UUID            `json:"client_id"`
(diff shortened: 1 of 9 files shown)


crypto.go
```diff
@@ -3,22 +3,51 @@ package e3dbClients
...
+}
+
 // Key wraps material generated using an algorithm/curve for use in cryptographic operations
 type Key struct {
 	Material string
 	Type     string // e.g. Curve25519
 }
 
+// Keys is a map of key type to key material
+type Keys map[string]string
```
emanb29: Should this just be a list of Keys? If not, what about adding types instead of string->string to indicate what those strings are used for?

emanb29: For rolling keys: We'd probably want a `map[KeyId or KeyNumber]Key` anyway. For keys of varying `Type`s, the information is already captured in `Key`

crypto.go
```diff
@@ -53,3 +92,121 @@ func GenerateSigningKeys() (SigningKeys, error) {
 	}
 	return signingKeys, err
 }
+
+// EncryptData accepts a record object of plain meta and plain data, returning
+// a record with data keys encrypted with dk and ak according to Tozstore spec.
```
emanb29: Is the spec something that could be linked here?

crypto.go
```diff
@@ -53,3 +92,121 @@ func GenerateSigningKeys() (SigningKeys, error) {
...
+	b := [NonceSize]byte{}
+	_, err := rand.Read(b[:])
+	if err != nil {
+		// we don't expect this to fail
+		panic("random number generation failed")
+	}
+	return &b
+}
+
+// BoxEncryptToBase64 uses asymmetric encryption key sot encrypt data
```
emanb29: `key sot` -> `keys to`

crypto.go
emanb29: Alternatively, if this is just a serialization format, should it be exported at all?

crypto.go
```diff
@@ -53,3 +80,121 @@ func GenerateSigningKeys() (SigningKeys, error) {
...
+
+// RandomNonce generates a random nonce of size NoneSize
+func RandomNonce() Nonce {
+	b := [NonceSize]byte{}
+	_, err := rand.Read(b[:])
+	if err != nil {
+		// we don't expect this to fail
+		panic("random number generation failed")
+	}
+	return &b
```
emanb29: Addressof calls are at different places in RandomNonce() and RandomSymmetricKey() -- is this intentional?

crypto.go
willmichael: Not sure if there is an internal spec outlined anywhere

crypto.go
willmichael: This future functionality is more for modes over rolling keys such as NIST vs Sodium mode; hence the map structure. We thought about adding types, but it adds too much overhead in creation/using the type that I think it's not worth it.

PR #59, opened 2019-08-12
Add notification client and basic happy/sad tests

emanb29:
Also sets the test emails to @example.com because `test.com` is a real company.


notificationClient/notificationClient.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type ToznyNotificationClient struct {
+	Host       string
+	authClient authClient.E3dbAuthClient
+}
+
+const (
+	// NotificationServiceBasePath is the base path used for notification-service calls
+	NotificationServiceBasePath = "v1/notification"
```
galxy25: take a look at the other clients, I think we're prefixing all these base paths with '/' with the thought that you should be able to join together any combination of hosts + {path1...pathN) without having to think about whether you need to add a trailing slash to the path.

notificationClient/notificationClient.go
```diff
@@ -0,0 +1,38 @@
...
+
+// CreateNotification pushes a notification via an internal endpoint. Requires the bootstrap client.
+func (nc *ToznyNotificationClient) CreateNotification(ctx context.Context, params CreateNotificationRequest) (*NotificationMeta, error) {
+	var result *NotificationMeta
+	path := nc.Host + "/internal/" + NotificationServiceBasePath + "/"
+	request, err := e3dbClients.CreateRequest(http.MethodPost, path, params)
+	if err != nil {
+		return result, e3dbClients.NewError(err.Error(), path, 0)
+	}
+	err = e3dbClients.MakeE3DBServiceCall(&nc.authClient, ctx, request, &result)
```
galxy25: In line with trying to push the greatest that is tsv1 , we should probably make this client authenticate it's calls with signature based auth (see storage v2/notes and identity clients for examples), it shouldn't be a change for the notification service itself as with either valid authentication method cyclops will construct a X-TOZNY-AUTHN header for the service to use as an auth check. (no need to hold off on dev deployments based on this change though)

notificationClient/notificationClient_test.go
```diff
@@ -0,0 +1,85 @@
...
+var (
+	e3dbAuthHost          = utils.MustGetenv("E3DB_AUTH_SERVICE_HOST")
+	e3dbAccountHost       = utils.MustGetenv("E3DB_ACCOUNT_SERVICE_HOST")
+	toznyNotificationHost = utils.MustGetenv("TOZNY_NOTIFICATION_SERVICE_HOST")
+	e3dbAPIKey            = utils.MustGetenv("E3DB_API_KEY_ID")
+	e3dbAPISecret         = utils.MustGetenv("E3DB_API_KEY_SECRET")
+	ValidClientConfig     = e3dbClients.ClientConfig{
+		APIKey:    e3dbAPIKey,
+		APISecret: e3dbAPISecret,
+		Host:      toznyNotificationHost,
```
galxy25: for all external endpoints a client should use CYCLOPS_SERVICE_HOST (this will ensure the request is authenticated and the auth n header will be set)

notificationClient/notificationClient.go
emanb29: Oops. Didn't see the "no need to hold off" part. Oh well. Added that authenticator to notification-service@working

notificationClient/notificationClient.go
emanb29: This was copied from Billing. I've changed it here, shall I change it there as well? (Not sure if other things use the BillingServiceBasePath variable since it's exposed)

galxy25: Yes please, looks like its not used by anything outside of e3db-clients-go https://github.com/search?l=&q=BillingServiceBasePath+user%3Atozny&type=Code

notificationClient/api.go
```diff
@@ -0,0 +1,28 @@
...
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const (
+	ChannelEmail = "email"
+)
+
+// CreateNotificationRequest is the api-level struct which represents an objcet expected by notification-service's POST / endpoint
```
efabens: ojcet -> object

PR #64, opened 2019-09-10
Update authClient to renew context as needed.

galxy25:


authClient/authClient.go
```diff
@@ -66,9 +67,15 @@ func (c *E3dbAuthClient) AuthenticateE3DBClient(ctx context.Context, token strin
 
 // AuthHTTPClient returns an http client that can be used for
 // making authenticated requests to an e3db endpoint using the provided context.
+// The first time AuthHTTPClient is called it use the provided context to create a new http client
+// and will reuse that context until it has expired
 func (c *E3dbAuthClient) AuthHTTPClient(ctx context.Context) *http.Client {
-	if c.httpClient == nil {
+	if c.httpClient == nil || c.ctx.Err() != nil {
+		if c.ctx != nil {
```
willmichael: Isn't the problem the context is expiring thus the http client we create from that context is invalid; wouldn't storing this context and creating a new httpClient from the same expired context lead to the same issue? 

Also I'm a fan of more logging, but I'm not sure what value this gives us here other than an extra log every hour which we can't turn off (or however long it takes to expire); I'd like to see a prefix string added if it stays in `AuthClient context expired: `

authClient/authClient.go
```diff
@@ -66,9 +67,15 @@ func (c *E3dbAuthClient) AuthenticateE3DBClient(ctx context.Context, token strin
 
 // AuthHTTPClient returns an http client that can be used for
 // making authenticated requests to an e3db endpoint using the provided context.
+// The first time AuthHTTPClient is called it use the provided context to create a new http client
+// and will reuse that context until it has expired
```
efabens: This still reuses contexts from requests for other requests, which seems off. Should this just use a background context instead?

authClient/authClient.go
willmichael: After re-reading, I amend my edit: We're not refreshing the context here we're just refreshing the client made from a possibly expired context. Wouldn't this lead to the same error?

galxy25:
@efabens @willmichael Sigh, context is too complicated. The original error is the first time we use the authClient's http client, it takes the context we give it and holds on to it. Eventually that context expires, and prior to this change we would never renew it. Now, when the original context has expired, it will get refreshed the next time the auth client is called

authClient/authClient.go
galxy25: Thats actually explicitly how we're using context here, the top level request sets the context and passes it down to every other function that is processing the request, allowing the request to be cancelled upstream and have any downstream work stop. Using a background context here would be some implicit magic that would break the contract expected by a user of this library(e.g. I give you context, you use context)

authClient/authClient.go
galxy25: No, this will fix the original error because now instead of the original context expiring and never being refreshed, it will.

willmichael: oh I see, so IF the stored context is expired, use the new context provided and store it; I would argue this is a level of magic caching, bc as a user I'd expect you to use the context I'm passing in but 🤷‍♀

willmichael: Why do we cache the context to begin with, is creating a httpClient from a context expensive?

galxy25: we cache the context to avoid calling auth every time we need to make a service to service call(the code got to its current state because the indexer was bringing auth down)

PR #73, opened 2019-10-03
Add bulk delete endpoint to storage

efabens:


storageClient/storageClient.go
```diff
@@ -74,12 +78,30 @@ func (c *StorageClient) Challenge(ctx context.Context, noteID string) ([]string,
...
 
+func (c *StorageClient) BulkDeleteByClient(ctx context.Context, clientID uuid.UUID, limit int) (BulkDeleteResponse, error) {
+	path := c.Host + "/internal" + storageServiceBasePath + "/notes/bulk/" + clientID.String()
+	request, err := e3dbClients.CreateRequest("DELETE", path, nil)
+	var resp BulkDeleteResponse
+	if err != nil {
+		return resp, err
+	}
+	urlParams := request.URL.Query()
+	if limit != 0 {
```
willmichael: do we need this check//do we need to also check for negative values?

efabens: We could, the services check and deal with negative values. This check is really more of a way to imply that you want to not attach the query param by putting zero. The other way to do that would be a pointer to the int and check for nil? pointers to ints just feels weird to me, but might be more appropriate.

storageClient/storageClient.go
```diff
@@ -74,12 +78,30 @@ func (c *StorageClient) Challenge(ctx context.Context, noteID string) ([]string,
...
+		urlParams.Set("limit", strconv.Itoa(limit))
+	}
+	request.URL.RawQuery = urlParams.Encode()
+	err = e3dbClients.MakeE3DBServiceCall(c.E3dbAuthClient, ctx, request, &resp)
```
galxy25: We've been slowly chipping away at killing old style auth, why is this internal endpoint using/need to use old style auth?

PR #76, opened 2019-10-10
Add identity client methods for hosted broker login flow.

galxy25:


identityClient/api.go
```diff
@@ -98,7 +98,35 @@ type RegisterRealmBrokerIdentityRequest struct {
...
+// BrokerChallengeRequest wraps Identifying information for the Identity initiating the challenge.
+type BrokerChallengeRequest struct {
+	RealmName string // The name of the realm the Identity is a member of.
+	Action    string `json:"action"`   // The requested broker flow action to perform. Currently only challenge is supported.
+	Username  string `json:"username"` // The name for the Identity initiating the challenge.
+}
+
+// EmailOTP wraps a one time password provided via an email challenge.
+type EmailOTP struct {
+	EmailOTP string // The one-time password from the email challenge issued.
```
lkwdwrd: does this need a `json:` tag?

identityClient/api.go
```diff
@@ -98,7 +98,35 @@ type RegisterRealmBrokerIdentityRequest struct {
...
+}
+
+// BrokerLoginRequest proof that the Identity has completed the broker challenge
+// along with key material to encrypt the login response.
+type BrokerLoginRequest struct {
+	RealmName    string            // The name of the realm the Identity is a member of.
+	Action       string            `json:"action"`                // The requested broker flow action to perform. Currently only login is supported
+	NoteID       uuid.UUID         `json:"note_id"`               // The ID of the recovery Note the email challenge was for.
+	PublicKeys   map[string]string `json:"public_key"`            // The public key to use to encrypt the recovery note.
+	SigningKeys  map[string]string `json:"signing_key,omitempty"` // The signing key to use to verify the integrity of the recovery note.
```
lkwdwrd: Because this is notes, this should never be empty -- any reason to `omitempty`?

galxy25: good catch, copy pasta

identityClient/api.go
galxy25: good catch

PR #84, opened 2019-11-19
Add body to note challenge

efabens:


storageClient/storageClient.go
```diff
@@ -62,9 +62,9 @@ func (c *StorageClient) ReadNote(ctx context.Context, noteID string, eacpParams
 	return result, err
 }
 
-func (c *StorageClient) Challenge(ctx context.Context, noteID string) ([]string, error) {
+func (c *StorageClient) Challenge(ctx context.Context, noteID string, body ChallengeRequestBody) ([]string, error) {
```
galxy25: nit: for consistency with other methods in this file and module this should be more like `params ChallenengeRequest`

PR #85, opened 2019-11-21
Add storageClient API definitions for TozID EACP 

galxy25:
- Update storageClient API definitions for activate challenge endpoint to include more fully specified activated challenges.
- Fix identity api struct semantic error.

### DIFF
--- a/storageClient/api.go
+++ b/storageClient/api.go
@@ -33,6 +33,7 @@ type EACP struct {
 	EmailEACP      *EmailEACP      `json:"email_eacp,omitempty"`
 	LastAccessEACP *LastAccessEACP `json:"last_access_eacp,omitempty"`
 	ToznyOTPEACP   *ToznyOTPEACP   `json:"tozny_otp_eacp,omitempty"`
+	TozIDEACP      *TozIDEACP      `json:"tozid_eacp,omitempty"`
 }
 
 type EmailEACP struct {
@@ -51,6 +52,10 @@ type LastAccessEACP struct {
 type ToznyOTPEACP struct {
 	Include bool `json:"include"`
 }
+
+// TozIDEACP wraps an EACP requiring the proxying of a one time password OTP
+// embedded as the nonce claim for a valid & signed TozID realm auth JWT token
+type TozIDEACP struct{}
 
 type BulkDeleteResponse struct {
 	ClientID    uuid.UUID `json:"client_id"`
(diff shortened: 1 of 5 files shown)


storageClient/storageClient_test.go
```diff
@@ -49,9 +48,31 @@ func TestFullSignatureAuthenticationFlowSucceedsWithUID(t *testing.T) {
...
 	}
-	_, err = StorageClient.WriteNote(ctx, *noteToWrite)
+	_, err = StorageClient.WriteNote(testCtx, *noteToWrite)
 	if err != nil {
-		t.Errorf("unable to write note %+v, to storage servive %s\n", noteToWrite, err)
+		t.Errorf("unable to write note %+v, to storage service %s\n", noteToWrite, err)
+	}
+}
+
+func TestAddTozIDEACPWhenCreatingNote(t *testing.T) {
```
efabens: What's a Toz Idea cpw 🤣

storageClient/api.go
```diff
@@ -80,9 +85,36 @@ type PrimeResponseBody struct {
...
+type ChallengeResponse struct {
+	EmailEACPChallenge *EmailEACPChallengeResponse `json:"email_eacp,omitempty"`
+	TozIDEACPChallenge *TozIDEACPChallengeResponse `json:"tozid_eacp,omitemtpy"`
+}
+
+type EmailEACPChallengeResponse struct {
+	EmailAddress string `json:"email_address"`
+	TemplateName string `json:"template"`
+	ProviderLink string `json:"provider_link"`
+	OTP          string `json:"otp"`
```
lkwdwrd: We definitely don't want to return the OTP or provider link from this EACP -- it should only be delivered via the email channel out of band of this request flow.

lkwdwrd: As mentioned in the API spec, I'm actually thinking we don't want to respond with any of this information. Happy to discuss, but in general this should be kept to a very minimum.

storageClient/api.go
```diff
@@ -80,9 +85,36 @@ type PrimeResponseBody struct {
...
+}
+
+// TozIDEACPChallengeRequest wraps parameters for an activated TozID EACP on a note.
+type TozIDEACPChallengeResponse struct {
+	ExpiresAt time.Time `json:"expires_at"`
+	OTP       string    `json:"otp"`
```
lkwdwrd: change name to something indicating this is a challenge value that must be signed over in an identity token as the nonce value.

PR #87, opened 2019-12-05
Note proxy

lkwdwrd:


main.go
```diff
@@ -95,6 +95,15 @@ func MakeSignedServiceCall(ctx context.Context, request *http.Request, keypair S
 	return err
 }
 
+// MakeProxiedSignedCall attempts to call an e3db service using the provided
```
lkwdwrd: TozStore plz

main.go
```diff
@@ -95,6 +95,15 @@ func MakeSignedServiceCall(ctx context.Context, request *http.Request, keypair S
...
+// signature to authenticate the request.
+func MakeProxiedSignedCall(ctx context.Context, headers http.Header, request *http.Request, result interface{}) error {
```
lkwdwrd: need headers when request is present?

lkwdwrd: this may not be needed, you can add the headers to the request in the implementing method instead. This is simpler, maybe a bit of code duplication, but clearer.

PR #90, opened 2019-12-30
Add internal delete note methods

efabens:


galxy25:
@efabens don't forget to update the api specs too ;-)

efabens:
I've got that one coming shortly as well

PR #96, opened 2020-01-30
Notification and Identity client calls

willmichael:


crypto.go
```diff
@@ -158,6 +158,24 @@ func RandomNonce() Nonce {
 	return &b
 }
 
+// GenerateRandomBytes generate a random number of bytes
+func GenerateRandomBytes(n int) []byte {
+	b := make([]byte, n)
+	_, err := rand.Read(b)
+	if err != nil {
+		// we don't expect this to fail
+		panic("random number generation failed")
```
efabens: I don't think we want this to panic

willmichael: Normally I would agree, but I'm following the above pattern of using panic when this happens to fail

lkwdwrd: I agree, I'd rather return an error than panic here

willmichael: hmm some notes in the golang docs refer to 
```
// Read generates len(p) random bytes and writes them into p. It
// always returns len(p) and a nil error.
```
indicating that it always returns a nil error simply to conform to the ioReader interface; however, I will update it in favor of returning error as this note is from the math package not the crypto/rand we're using

PR #98, opened 2020-03-19
Add changes to identity spec and crypto method to aid in testing

willmichael:


identityClient/api.go
```diff
@@ -177,14 +177,16 @@ type InitiateUserChallengeResponse struct {
 }
 
 type InitiateRegisterDeviceRequest struct {
-	TempPublicKey string `json:"temporary_public_key"`
+	TempPublicKey           string `json:"temporary_public_key"`
+	TempPublicEncryptionKey string `json:"temporary_public_encryption_key"`
```
lkwdwrd: I realize this is adding to an existing construct, but typically when public signing and encryption key are sent to our API the encryption public key is referred to as 'public key' and the singing public key is referred to as 'signing key'. This flips that on it's head a little, but for consistency it may be worth considering the internal naming even if the JSON keys are different. A little easier to quickly understand what's happening for those familiar with our code.

identityClient/api.go
```diff
@@ -196,6 +198,7 @@ type CompleteUserDeviceRegisterRequest struct {
 	DeviceID        string `json:"device_id"`         // Unique device identifier
 	DeviceName      string `json:"device_name"`       // Human readable device name
 	DevicePublicKey string `json:"device_public_key"` // PublicKey to permanently stored PrivateKey on device
+	TOTP            string `json:"secret"`
```
lkwdwrd: This key is not the actual secret, right? We're sending the 6-digit TOTP?

willmichael: Right this is misleading naming that should be corrected, it's the 6 digit totp for verification.

identityClient/api.go
willmichael: Yeah its a lack of foresight by me unfortunately, keeping consistent naming would require 3 more changes in frontend space (mobile apps and browser) as it's a breaking api change; since we won't have frontend support for totp at this time I don't think it a change is necessary and can come with those frontend updates.

PR #105, opened 2020-08-18
Makes realm names in keycloak paths lowercase

efabens:
- Fixes singing key generation bug as well

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -70,7 +70,7 @@ type PublicSigningKeys Keys
 // returning the signing keys and error (if any).
 func GenerateSigningKeys() (SigningKeys, error) {
 	var signingKeys SigningKeys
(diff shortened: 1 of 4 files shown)


crypto.go
```diff
@@ -70,7 +70,7 @@ type PublicSigningKeys Keys
 // returning the signing keys and error (if any).
 func GenerateSigningKeys() (SigningKeys, error) {
 	var signingKeys SigningKeys
-	signingKeyBytes, privateSigningKeyBytes, err := box.GenerateKey(rand.Reader)
+	signingKeyBytes, privateSigningKeyBytes, err := sign.GenerateKey(rand.Reader)
```
willmichael, line 73: would any consuming services have problems if they used this method?

efabens: this was a bug that just got added. This fixes the bug

PR #106, opened 2020-08-18
Add internal ldap cache endpoints to identity client

lkwdwrd:
Adds methods for interacting with the new LDAP cache endpoint in identity service.

### DIFF
--- a/identityClient/api.go
+++ b/identityClient/api.go
@@ -208,14 +208,24 @@ type CompleteChallengeRequest struct {
 	SignedTime      int64  `json:"sign_time"`        // time challenge was signed
 }
 
+// LDAPCache wraps the LDAP JSON data passed between services for caching
 type LDAPCache struct {
-	ID                 string              `json:"id,pk"`
-	LDAPID             string              `json:"ldap_id"`
-	DN                 string              `json:"dn"`
-	RdnAttributeName   string              `json:"rdn_attribute_name"`
-	Classes            []string            `json:"classes"`
-	Attributes         map[string][]string `json:"attributes"`
-	ReadOnlyAttributes []string            `json:"read_only_attributes"`
-	Groups             []string            `json:"groups"`
-	Roles              []string            `json:"roles"`
+	// User's Keycloak ID
+	ID string `json:"id,pk"`
+	// User's LDAP UUID (assigned by the LDAP server)
+	LDAPID string `json:"ldap_id"`
+	// The domain query used to find the user in the LDAP server
+	DN string `json:"dn"`
+	// Relative Distinguished Name addresses this specific user relative to the DN
+	RdnAttributeName string `json:"rdn_attribute_name"`
+	// Used to define what type of LDAP object this is
+	Classes []string `json:"classes"`
+	// Information about the LDAP object in key -> multi-value pairs
+	Attributes map[string][]string `json:"attributes"`
+	// Attribute keys which should not be writable
+	ReadOnlyAttributes []string `json:"read_only_attributes"`
+	// A set of Keycloak group IDs the user had at the time of caching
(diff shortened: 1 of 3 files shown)


identityClient/api.go
```diff
@@ -207,3 +207,15 @@ type CompleteChallengeRequest struct {
 	SignedChallenge string `json:"signed_challenge"` // signed challenge in the form of "{challenge}{timestamp}"
 	SignedTime      int64  `json:"sign_time"`        // time challenge was signed
 }
+
+type LDAPCache struct {
```
galxy25, line 211: comments for any public module exports please, adds documentation for us and future users (especially now that e3db-clients-go is fully exposed via e3db-go)

identityClient/api.go
```diff
@@ -207,3 +207,15 @@ type CompleteChallengeRequest struct {
...
+	ID                 string              `json:"id,pk"`
+	LDAPID             string              `json:"ldap_id"`
+	DN                 string              `json:"dn"`
+	RdnAttributeName   string              `json:"rdn_attribute_name"`
```
galxy25, line 215: and comments for struct fields that are not apparent to non LDAP experts. If possible for struct field names try to not go with contractions (obviously the json tag has to for compatibility with api's, but this is code we want to be compatible or at least more readable for humans)

identityClient/identityClient_test.go
```diff
@@ -13,15 +13,15 @@ import (
...
-	toznyCyclopsHost            = os.Getenv("TOZNY_CYCLOPS_SERVICE_HOST")
-	internalIdentityServiceHost = os.Getenv("E3DB_IDENTITY_SERVICE_HOST")
-	e3dbAuthHost                = os.Getenv("E3DB_AUTH_SERVICE_HOST")
-	e3dbAccountHost             = os.Getenv("E3DB_ACCOUNT_SERVICE_HOST")
-	e3dbAPIKey                  = os.Getenv("E3DB_API_KEY_ID")
-	e3dbAPISecret               = os.Getenv("E3DB_API_KEY_SECRET")
-	e3dbClientID                = os.Getenv("E3DB_CLIENT_ID")
-	bootstrapPublicSigningKey   = os.Getenv("BOOTSTRAP_CLIENT_PUBLIC_SIGNING_KEY")
-	bootstrapPrivateSigningKey  = os.Getenv("BOOTSTRAP_CLIENT_PRIVATE_SIGNING_KEY")
+	toznyCyclopsHost            = utils.MustGetenv("TOZNY_CYCLOPS_SERVICE_HOST")
```
galxy25, line 16: imo, unlike production services, tests don't need to fail for missing environment variables. For something like this which is only ran by developers, I'd prefer leaving it open to being able to only specify in your local environment the variables needed for one's own testing.

lkwdwrd: I actually hit an error where tests would simply not run because of missing environment variables. I'm OK having them optional if tests don't break, but it took a bit of work to figure out what was actually missing. I'm fine dropping back to `os.` level.

identityClient/api.go
lkwdwrd: Will do

identityClient/api.go
lkwdwrd: I actually don't know what many of these fields are... they are names from the Java LDAP objects, I just cached the values with the same names used... I _think_ dn is 'domain name' for example. I'll see if some quick googling will reveal what these things are.

identityClient/identityClient_test.go
galxy25: welp, I suppose better the devil you know. Local developers can always comment out what ones they don't need, or when adding new ones in the future optionally make it a MustGet

PR #110, opened 2020-08-28
Add struct fields for client side storing and serializing of Identity secret key material.

galxy25:


identityClient/api.go
```diff
@@ -38,17 +38,19 @@ type ListRealmsResponse struct {
...
+	ID                    int64             `json:"id"`
+	ToznyID               uuid.UUID         `json:"tozny_id"` // Tozny Client ID
+	RealmID               int64             `json:"realm_id"`
+	RealmName             string            `json:"realm_name"`
+	Name                  string            `json:"name"`
+	FirstName             string            `json:"first_name"`
+	LastName              string            `json:"last_name"`
+	APIKeyID              string            `json:"api_key_id"`
+	APIKeySecret          string            `json:"api_secret_key"`
+	PublicKeys            map[string]string `json:"public_key"`
```
willmichael, line 50: thoughts on plurality here? publicKeys vs public_key

galxy25: its public keys because you can have multiple public keys (e.g. one for FIPS/Nist curves and one for other curves) and the key is only `public_key` because it has to conform to the badly pluralized json fields for (originally account) and now (client service)

PR #121, opened 2020-09-29
Fetch of identity info using the list and details endpoints

lkwdwrd:


efabens (approved):

go.mod
```diff
@@ -3,11 +3,10 @@ module github.com/tozny/e3db-clients-go
 require (
 	github.com/golang/protobuf v1.3.1 // indirect
 	github.com/google/uuid v1.1.0
+	github.com/tozny/keycloak-client v0.0.7
```
galxy25, line 6: is this being used? It shouldn't need to as identity management service is wrapping all the Keycloak endpoints we need to call right?

galxy25 (approved):

go.mod
lkwdwrd: no -- it should not be used. Will double check, this should be pulled out

lkwdwrd: Oh, I think I was in the wrong repo when I typed `go get...`

PR #128, opened 2020-10-30
Adds delete identites by realm-provider pair

efabens:


identityClient/identityClient.go
```diff
@@ -516,6 +517,16 @@ func (c *E3dbIdentityClient) InternalDeleteIdentity(ctx context.Context, realmNa
 	return err
 }
 
+func (c *E3dbIdentityClient) InternalDeleteIdentitiesByProvider(ctx context.Context, realmName string, providerID uuid.UUID) error {
```
galxy25, line 520: prefer adding a struct vs an additional param, if only for OCD/minalimiim of having to not import the uuid library :-P. Clean and lean

galxy25 (approved):

identityClient/identityClient.go
efabens: I'm lodging a formal protest because I think the uuid library should be imported everywhere just in case you might need it, it is a travesty it isn't just baked into the standard library 😛. but ok

lkwdwrd (approved):

PR #129, opened 2020-11-01
Add client methods, types and test for CRD on application protocol mappers.

galxy25:


identityClient/api.go
```diff
@@ -588,8 +607,78 @@ type FetchApplicationSAMLDescriptionRequest struct {
 	Format        string
 }
 
+// ApplicationMapper wraps related values related to an application mapper for auth flows of a given protocol.
+type ApplicationMapper struct {
+	// The UUID assigned to the application mapper
+	ID string `json:"id"`
```
efabens, line 613: It probably makes sense to have uuids be uuids instead of strings

efabens: There are probably others throughout.

efabens (approved):

PR #131, opened 2020-11-19
changes for create group endpoint

rosasck:


storageClient/api.go
```diff
@@ -29,6 +29,16 @@ type Note struct {
 	Expires             bool              `json:"expires,omitempty"`
 }
 
+//Group wraps values related to managing client membership and capabilities for resources such as records.
+type Group struct {
```
willmichael, line 33: You should separate the difference between a GroupRequest and a GroupResponse -- when we make the request we don't want to have to provide GroupID/CreatedAt/LastModified/AccountID, but we want all those fields when the result is returned to us

storageClient/storageClient.go
```diff
@@ -45,6 +45,17 @@ func (c *StorageClient) WriteNote(ctx context.Context, params Note) (*Note, erro
 	return result, err
 }
 
+func (c *StorageClient) PostGroups(ctx context.Context, params Group) (*Group, error) {
```
willmichael, line 48: We should have a more functional name like "CreateGroup" and accept the different param/response types as mentioned above

storageClient/api.go
```diff
@@ -29,6 +29,18 @@ type Note struct {
 	Expires             bool              `json:"expires,omitempty"`
 }
 
+type GroupRequest struct {
+	Name      string `json:"group_name"`
+	PublicKey string `json:"public_key"`
+}
+type GroupRespond struct {
```
efabens, line 36: I think we use `Response` usually not `Respond`

efabens (approved):

PR #134, opened 2020-11-23
Changes for Read Group Endpoint

rosasck:
Added Function for Read Group Endpoint, Plus a change from previous PR that didn't get fixed.

### DIFF
--- a/storageClient/storageClient.go
+++ b/storageClient/storageClient.go
@@ -46,7 +46,7 @@ func (c *StorageClient) WriteNote(ctx context.Context, params Note) (*Note, erro
 }
 
 // CreateGroup creates a group using the specified parameters,
-//returning the created group and error (if any)
+// returning the created group and error (if any).
 func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupRequest) (*Group, error) {
 	var result *Group
 	path := c.Host + storageServiceBasePath + "/groups"
@@ -54,6 +54,20 @@ func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupReque
 	if err != nil {
 		return result, err
 	}
+	err = e3dbClients.MakeSignedServiceCall(ctx, request, c.SigningKeys, c.ClientID, &result)
+	return result, err
+}
+
+// DescribeGroup fetches a group using the specified parameters,
+// returning the group and error (if any).
+func (c *StorageClient) DescribeGroup(ctx context.Context, params DescribeGroupRequest) (*Group, error) {
+	var result *Group
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String()
+	request, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+
 	err = e3dbClients.MakeSignedServiceCall(ctx, request, c.SigningKeys, c.ClientID, &result)
 	return result, err
 }
--- a/storageClient/api.go
+++ b/storageClient/api.go
@@ -44,6 +44,11 @@ type Group struct {
 	CreatedAt    time.Time `json:"created_at"`
 	LastModified time.Time `json:"last_modified"`
 }
+
+// DescribeGroupRequest wraps values used to look up Groups
+type DescribeGroupRequest struct {
+	GroupID uuid.UUID `json:"group_id"`
+}
 type EACP struct {
 	EmailEACP      *EmailEACP      `json:"email_eacp,omitempty"`
 	LastAccessEACP *LastAccessEACP `json:"last_access_eacp,omitempty"`


storageClient/storageClient.go
```diff
@@ -58,6 +58,25 @@ func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupReque
 	return result, err
 }
 
+// getGroupByGroupNameAndAccountID fetches a group using the specified parameters,
+// returning the group and error (if any).
+func (c *StorageClient) getGroupByGroupNameAndAccountID(ctx context.Context, groupName string, accountID string) (*Group, error) {
```
galxy25, line 63: https://github.com/tozny/e3db-clients-go/blob/master/identityClient/identityClient.go#L65
this should be named at a level one abstraction layer higher then the http method and params, e.g. DescribeGroup and collect the group name and account id individual params into a DescribeGroupRequest struct

storageClient/storageClient.go
```diff
@@ -58,6 +58,25 @@ func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupReque
...
+	var result *Group
+	path := c.Host + storageServiceBasePath + "/groups/" + groupName
+
+	request, err := e3dbClients.CreateRequest("GET", path, nil)
+	if err != nil {
+		return result, err
+	}
+	urlParams := request.URL.Query()
+	urlParams.Set("account_id", accountID)
+	urlParams.Set("group_name", groupName)
```
galxy25, line 73: group name should be in the path vs query, kinda like this
https://github.com/tozny/e3db-clients-go/blob/master/identityClient/identityClient.go#L661
except you need to url encode the group name as it could have spaces

galxy25: https://www.urlencoder.io/golang/#url-decoding-a-path-segment-in-golang

galxy25 (changes requested):
See comments

storageClient/storageClient.go
rosasck: Do I still keep group name here to be encoded?

willmichael: @galxy25 I think if a param needs to be url encoded it should live within the query params rather than the path

efabens: I think I would side with levi here on keep it in the path. it allows for more expressive future endpoints

efabens: But we might want to restrict (in code) what groupnames can be

rosasck:
```
func (c *StorageClient) DescribeGroup(ctx context.Context, params DescribeGroupRequest) (*Group, error) {
	var result *Group
        groupNameURL:= url.Parse(fmt.Sprintf("/%s", params.Name))
	path := c.Host + storageServiceBasePath + "/groups/" + groupNameURL
	request, err := e3dbClients.CreateRequest("GET", path, params)
	if err != nil {
		return result, err
	}
	urlParams := request.URL.Query()
	urlParams.Set("account_id", params.AccountID.String())
	request.URL.RawQuery = urlParams.Encode()
	err = e3dbClients.MakeSignedServiceCall(ctx, request, c.SigningKeys, c.ClientID, &result)
	return result, err
}
```

rosasck: Would the encoding part look like this?

storageClient/storageClient.go
```diff
@@ -58,6 +59,24 @@ func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupReque
...
+func (c *StorageClient) DescribeGroup(ctx context.Context, params DescribeGroupRequest) (*Group, error) {
+	var result *Group
+	path := c.Host + storageServiceBasePath + "/groups/" + fmt.Sprintf("/%s", params.Name)
+	request, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	urlParams := request.URL.Query()
+	urlParams.Set("account_id", params.AccountID.String())
+	urlParams.Set("group_name", params.Name)
```
galxy25, line 73: name only needs to be set in the path

storageClient/storageClient.go
galxy25: @rosasck use https://golang.org/src/net/url/url.go?s=7851:7884#L264 QueryEscape to encode the params.name as part of the path. Use on the server side when parsing this value from the path https://golang.org/pkg/net/url/#QueryUnescape

storageClient/storageClient.go
```diff
@@ -58,6 +58,20 @@ func (c *StorageClient) CreateGroup(ctx context.Context, params CreateGroupReque
 	return result, err
 }
 
+// DescribeGroup fetches a group using the specified parameters,
+// returning the group and error (if any).
+func (c *StorageClient) DescribeGroup(ctx context.Context, params DescribeGroupRequest) (*Group, error) {
+	var result *Group
+	path := c.Host + storageServiceBasePath + "/groups"
```
galxy25, line 65: + "/" + params.GroupID

galxy25 (approved):

galxy25 (approved):

storageClient/storageClient.go
rosasck: it would be params.Groupid.String() right?

efabens (approved):

PR #139, opened 2020-11-26
Added accountClient v2 health and service checks

carrala:
Eli & I decided to add HostV2 to the E3dbAccoundClient so that these tests can be done, but I would like feedback on how other people think this should be done.
We're thinking there are three options:
1. 1 client with hosts for v1 and v2
2. 2 different clients with 1 host each
3. 1 client with 1 host where checks are managed by Cyclops

### DIFF
--- a/accountClient/accountClient.go
+++ b/accountClient/accountClient.go
@@ -11,11 +11,12 @@ import (
 	"github.com/tozny/e3db-clients-go/authClient"
 )
 
+// HTTP PATH prefix for calls to the e3db Account service for v1
 const (
-	AccountServiceBasePath = "v1/account" //HTTP PATH prefix for calls to the e3db Account service
+	AccountServiceBasePath = "v1/account"
 )
 
-//E3dbAccountClient implements an http client for communication with an e3db Account service.
+// E3dbAccountClient implements an http client for communication with an e3db Account service.
 type E3dbAccountClient struct {
 	APIKey    string
 	APISecret string
@@ -80,7 +81,7 @@ func (c *E3dbAccountClient) RegisterClient(ctx context.Context, params ClientReg
 	return result, err
 }
 
-// RegisterClient registers a client via a proxied call to client service by the account service.
+// ProxyiedRegisterClient registers a client via a proxied call to client service by the account service.
 // This method is intended for TESTING the functionality of the integrated client service. Not intended for future use.
 func (c *E3dbAccountClient) ProxyiedRegisterClient(ctx context.Context, params ProxiedClientRegistrationRequest) (*ProxiedClientRegisterationResponse, error) {
 	var result *ProxiedClientRegisterationResponse
--- a/accountClient/accountClientV2.go
+++ b/accountClient/accountClientV2.go
@@ -1,0 +1,56 @@
+package accountClient
+
+import (
+	"context"
+
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
+)
+
+// HTTP PATH prefix for calls to the e3db Account service for v2
+const (
+	AccountServiceV2BasePath = "v2/account"
+)
+
+// E3dbAccountClient implements an http client for communication with an e3db Account service.
+type E3dbAccountClientV2 struct {
+	APIKey    string
+	APISecret string
+	Host      string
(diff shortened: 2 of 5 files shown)


efabens:
Option three is that the host must be cyclops so that it does the routing. Slightly cleaner but causes potential minor backwards incompatibility issues and means you can't hit both services directly if there is a want to do that

efabens (approved):
Assuming folks are happy with the configuration change

galxy25 (changes requested):
Refactor to either 1 account client, code should be able to create v1 and v2 clients but with a different host, e.g.

```go
accountServiceV1Client := accountClient.New({config with v1 host goes here})
accountServiceV2Client  :=  accountClient.New({config with v2 host goes here})
```

or 2 clients (basically create a new client in a file called accountClientV2.go

efabens:
Let's do two files/cleints then. Because just specifying different hosts for the same methods has the potential for unexpected outcomes and lots of 404s if you don't realize

accountClient/accountClientV2.go
```diff
@@ -0,0 +1,56 @@
...
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
+)
+
+//HTTP PATH prefix for calls to the e3db Account service for v1 and v2
+const (
+	AccountServiceV2BasePath = "v2/account"
+)
+
+//E3dbAccountClient implements an http client for communication with an e3db Account service.
```
efabens, line 15: Space between // and text

accountClient/accountClientV2.go
```diff
@@ -0,0 +1,56 @@
+package accountClient
+
+import (
+	"context"
+
+	e3dbClients "github.com/tozny/e3db-clients-go"
+	"github.com/tozny/e3db-clients-go/authClient"
+)
+
+//HTTP PATH prefix for calls to the e3db Account service for v1 and v2
```
efabens, line 10: This should just be v2

accountClient/accountClient.go
```diff
@@ -11,8 +11,9 @@ import (
 	"github.com/tozny/e3db-clients-go/authClient"
 )
 
+//HTTP PATH prefix for calls to the e3db Account service for v1 and v2
```
efabens, line 14: this should just be v1

efabens (commented):
Couple of very minor nits

accountClient/accountClient.go
```diff
@@ -80,7 +81,7 @@ func (c *E3dbAccountClient) RegisterClient(ctx context.Context, params ClientReg
 	return result, err
 }
 
-// RegisterClient registers a client via a proxied call to client service by the account service.
+// ProxyiedRegisterClient registers a client via a proxied call to client service by the account service.
```
galxy25, line 84: Thank you

accountClient/accountClientV2.go
```diff
@@ -0,0 +1,56 @@
...
+type E3dbAccountClientV2 struct {
+	APIKey    string
+	APISecret string
+	Host      string
+	*authClient.E3dbAuthClient
+}
+
+// ServiceCheckV2 checks whether the account service V2 is up and working.
+// returning error if unable to connect service
+func (c *E3dbAccountClientV2) ServiceCheckV2(ctx context.Context) error {
```
galxy25, line 25: no V2 suffix needed for these methods now that they're part of a separate client

accountClient/accountClient_test.go
```diff
@@ -14,15 +14,21 @@ import (
...
 	e3dbAPIKey        = os.Getenv("E3DB_API_KEY_ID")
 	e3dbAPISecret     = os.Getenv("E3DB_API_KEY_SECRET")
-	e3dbClientID      = os.Getenv("E3DB_CLIENT_ID")
 	ValidClientConfig = e3dbClients.ClientConfig{
 		APIKey:    e3dbAPIKey,
 		APISecret: e3dbAPISecret,
 		Host:      e3dbAccountHost,
 		AuthNHost: e3dbAuthHost,
 	}
+	ValidClientConfigV2 = e3dbClients.ClientConfig{
```
galxy25, line 26: Is this still needed?

galxy25 (approved):

PR #142, opened 2020-12-04
Feature/plat498 add group members

rosasck:
I based this branch off of Feature/PLAT505 List Groups since i would need things from that branch... let me know if this is wrong. I see the commit from that branch...


rosasck:
Need to fix

PR #143, opened 2020-12-07
Feature/plat498 add group members

rosasck:


storageClient/api.go
```diff
@@ -73,12 +73,22 @@ type AddingCapabilityRequest struct {
...
 
-// AddingClientToGroupRequest wraps values used to add a client to a Group.
-type AddingClientToGroupRequest struct {
-	ClientID       uuid.UUID `json:"client_id"`
-	Name           string    `json:"group_name"`
-	CapabilityName string    `json:"capability_name"`
-	AccountID      uuid.UUID `json:"account_id"`
+// NewMember wraps values needed to add a client to a Group.
+type NewMember struct {
+	ClientID       []uuid.UUID `json:"client_ids"`
```
efabens, line 78: When will a member have multiple client_ids?

rosasck: oops! missed that one. It was when i originally wasnt doing GroupMember and the Group member array

storageClient/api.go
```diff
@@ -73,12 +73,22 @@ type AddingCapabilityRequest struct {
...
+// NewMember wraps values needed to add a client to a Group.
+type NewMember struct {
+	ClientID       uuid.UUID `json:"client_ids"`
+	CapabilityName []string  `json:"capability_names"`
+}
+
+// AddToGroupRequest wraps the information of all members being added
+type AddToGroupRequest struct {
+	GroupID         uuid.UUID `json:"group_id"`
+	NewGroupMembers []NewMember
```
galxy25, line 85: this key should be `GroupMembers`

storageClient/api.go
```diff
@@ -73,12 +73,22 @@ type AddingCapabilityRequest struct {
...
 }
 
-// AddingClientToGroupRequest wraps values used to add a client to a Group.
-type AddingClientToGroupRequest struct {
-	ClientID       uuid.UUID `json:"client_id"`
-	Name           string    `json:"group_name"`
-	CapabilityName string    `json:"capability_name"`
-	AccountID      uuid.UUID `json:"account_id"`
+// NewMember wraps values needed to add a client to a Group.
+type NewMember struct {
```
galxy25, line 77: `GroupMember`

galxy25: `Wraps values for a member of a group`

galxy25 (changes requested):
requested changes in comments

rosasck:
After talking with @lkwdwrd, I did a draft of the membership key endpoint for clients

storageClient/api.go
```diff
@@ -31,9 +31,10 @@ type Note struct {
 
 // CreateGroupRequest wraps parameters needed to request creation of a group
 type CreateGroupRequest struct {
-	Name              string `json:"group_name"`
-	PublicKey         string `json:"public_key"`
-	EncryptedGroupKey string `json:"encrypted_group_key"`
+	Name                   string `json:"group_name"`
+	PublicKey              string `json:"public_key"`
+	EncryptedGroupKey      string `json:"encrypted_group_key"`
+	EncryptedMembershipKey string `json:"encrypted_membership_key"`
```
galxy25, line 37: I thought the group and the membership key are the same thing? If not and we need this reminder to update the api-spec accordingly.

rosasck: You are correct! I got in the mindset of "must add membership key to this create group"

rosasck:
Added Group membership key end point and test!

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,67 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+		GroupAdminID:      queenClient.ClientID,
+		NewMemberID:       groupMemberToAdd.ClientID,
+		EncryptedGroupKey: response.EncryptedGroupKey,
+		ShareePublicKey:   queenClient.EncryptionKeys.Public.Material,
+	}
+	membershipKeyResponse, err := queenClient.CreateGroupMembershipKey(testCtx, membershipKeyRequest)
+	if err != nil {
+		t.Fatalf("Failed to create membership key \n response %+v \n error %+v", membershipKeyResponse, err)
+	}
+	// Verify that the group membership key for the client matches the output for the group raw privat key above
```
galxy25, line 616: need to check the two unchecked errs on 617 & 618, and do this verification described in the comment

storageClient/api.go
```diff
@@ -73,12 +73,31 @@ type AddingCapabilityRequest struct {
...
+
+// AddToGroupRequest wraps the information of all members being added
+type AddToGroupRequest struct {
+	GroupID      uuid.UUID `json:"group_id"`
+	GroupMembers []GroupMember
+}
+
+// AddToGroupResponse wraps values for clients added to a Group.
+type AddToGroupResponse struct {
+	GroupID      uuid.UUID `json:"group_id"`
```
galxy25, line 91: Because the client can't add members to a group without knowing the GroupID, I don't think we need to return the GroupID to them in this response. For AddGroupMembers and RemoveGroupMembers the response can be a simple slice of GroupMembers

storageClient/api.go
```diff
@@ -73,12 +73,31 @@ type AddingCapabilityRequest struct {
...
-	AccountID      uuid.UUID `json:"account_id"`
+// NewMember wraps values needed to add a client to a Group.
+type GroupMember struct {
+	ClientID       uuid.UUID `json:"client_ids"`
+	MembershipKey  string    `json:"membership_key"`
+	CapabilityName []string  `json:"capability_names"`
+}
+
+// AddToGroupRequest wraps the information of all members being added
+type AddToGroupRequest struct {
```
galxy25, line 84: `AddGroupMembersRequest`

storageClient/api.go
```diff
@@ -73,12 +73,31 @@ type AddingCapabilityRequest struct {
 	AccountID      uuid.UUID `json:"account_id"`
 }
 
-// AddingClientToGroupRequest wraps values used to add a client to a Group.
-type AddingClientToGroupRequest struct {
-	ClientID       uuid.UUID `json:"client_id"`
-	Name           string    `json:"group_name"`
-	CapabilityName string    `json:"capability_name"`
-	AccountID      uuid.UUID `json:"account_id"`
+// NewMember wraps values needed to add a client to a Group.
```
galxy25, line 76: comment needs to be updated for the struct name below and to generalize this to `GroupMembers wraps values for a client that is a member of a group ` as this struct is used and returned as part of adding/removing and listing group members.

storageClient/storageClient.go
```diff
@@ -102,6 +104,18 @@ func (c *StorageClient) ListGroups(ctx context.Context, params ListGroupsRequest
 	return result, err
 }
 
+// AddMembersToGroup Adds Clients to a group and returns the succesfuly added clients (if any).
+func (c *StorageClient) AddMembersToGroup(ctx context.Context, params AddToGroupRequest) (*AddToGroupResponse, error) {
```
galxy25, line 108: `AddGroupMembers` for the method name please, `AddGroupMembersRequest` for the params, and `*[]GroupMember` for the first return value

storageClient/api.go
```diff
@@ -73,12 +73,31 @@ type AddingCapabilityRequest struct {
...
+type GroupMember struct {
+	ClientID       uuid.UUID `json:"client_ids"`
+	MembershipKey  string    `json:"membership_key"`
+	CapabilityName []string  `json:"capability_names"`
+}
+
+// AddToGroupRequest wraps the information of all members being added
+type AddToGroupRequest struct {
+	GroupID      uuid.UUID `json:"group_id"`
+	GroupMembers []GroupMember
```
galxy25, line 86: this needs a json struct tag

crypto.go
```diff
@@ -158,6 +158,19 @@ func DecryptData(Data map[string]string, ak SymmetricKey) (*map[string]string, e
 	return &decryptedData, nil
 }
 
+// EncryptPrivateKey Encrypts a private key using a keypair.
+func EncryptPrivateKey(privateKey Key, encryptionKeys EncryptionKeys) (string, error) {
+	rawPrivateKeyBytes, err := Base64Decode(privateKey.Material)
```
lkwdwrd, line 163: just to clarify... these are abstracted methods that use Base64URL encoding, right?

storageClient/api.go
```diff
@@ -73,12 +73,25 @@ type AddingCapabilityRequest struct {
...
-type AddingClientToGroupRequest struct {
-	ClientID       uuid.UUID `json:"client_id"`
-	Name           string    `json:"group_name"`
-	CapabilityName string    `json:"capability_name"`
-	AccountID      uuid.UUID `json:"account_id"`
+// GroupMember wraps values for a client that is a member of a group.
+type GroupMember struct {
+	ClientID       uuid.UUID `json:"client_ids"`
+	MembershipKey  string    `json:"membership_key"`
+	CapabilityName []string  `json:"capability_names"`
```
lkwdwrd, line 80: nit: plural since this is a list?

storageClient/storageClient.go
```diff
@@ -102,6 +104,18 @@ func (c *StorageClient) ListGroups(ctx context.Context, params ListGroupsRequest
 	return result, err
 }
 
+// AddGroupMembers Adds Clients to a group and returns the succesfuly added clients (if any).
+func (c *StorageClient) AddGroupMembers(ctx context.Context, params AddGroupMembersRequest) (*[]GroupMember, error) {
+	var result *[]GroupMember
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String() + "members"
```
lkwdwrd, line 110: is this missing a slash between params.GroupID and "members"?

storageClient/storageClient.go
```diff
@@ -113,6 +127,60 @@ func (c *StorageClient) UpsertNoteByIDString(ctx context.Context, params Note) (
...
+	// encrypted Group Key and the public key of
+	// the group admin specified in params,
+	// create an encrypted membership key that
+	// the new member can decrypt
+	wrappedEncryptedMembershipKey, err = c.WrapMembershipKeyForGroupMember(ctx, groupKey, params.NewMemberID)
+	return wrappedEncryptedMembershipKey, err
+}
+
+// WrapMembershipKeyForGroupMember tales the group key and wraps it for the new client
+func (c *StorageClient) WrapMembershipKeyForGroupMember(ctx context.Context, groupKey e3dbClients.SymmetricKey, newMemberID string) (string, error) {
```
lkwdwrd, line 152: Do not change if this is inline with the naming for other keys!

In documentation, we typically refer to things like encyrpted access key. In the design I noted the name for this as 'encrypted membership key' to stay consistent. In most SDK these method names are things like `encryptMembershipKeyForGroupMember()`. I'd prefer to use 'encrypted' instead of 'wrapped'.

Noting again -- if the other methods for keys in clients go use 'wrapped' -- stick with that. I'd rather this code stay consistent internally.

crypto.go
rosasck: Yes this is used to encrypt the private key for groups

storageClient/storageClient.go
```diff
@@ -113,6 +127,60 @@ func (c *StorageClient) UpsertNoteByIDString(ctx context.Context, params Note) (
...
+	// encrypted Group Key and the public key of
+	// the group admin specified in params,
+	// create an encrypted membership key that
+	// the new member can decrypt
+	wrappedEncryptedMembershipKey, err = c.encryptMembershipKeyForGroupMember(ctx, groupKey, params.NewMemberID)
+	return wrappedEncryptedMembershipKey, err
+}
+
+// WrapMembershipKeyForGroupMember tales the group key and wraps it for the new client
+func (c *StorageClient) encryptMembershipKeyForGroupMember(ctx context.Context, groupKey e3dbClients.SymmetricKey, newMemberID string) (string, error) {
```
galxy25, line 152: comment needs to be updated for the new method name and please make this a public method `Encrypt....` so it can be used by other libraries.

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,74 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
```
galxy25, line 616: what step below verifies the above comment?

galxy25 (approved):

crypto.go
galxy25: Only pushing on this because I think @rosasck might find the source of this answer interesting, but what question do you think Luke asked you / why do you think he asked?

lkwdwrd (approved):

rosasck:
Copy Pasta happy path tests

storageClient/storageClient_test.go
galxy25: this comment keeps getting resolved but I'm not seeing the verification implemented.

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Client to add to group Private key was not decoded correctly \n Client: %+v, err: %+v ", groupMemberToAdd, err)
+	}
+	groupKey, err := e3dbClients.DecryptEAK(membershipKeyResponse, queenClient.EncryptionKeys.Public.Material, rawEncryptionKey)
+	if err != nil {
+		t.Fatalf("The membership key was not decrypted correctly: MembershipKey: %+v, queenClient : %+v, group member to add private key: %+v ", membershipKeyResponse, queenClient.EncryptionKeys.Public.Material, rawEncryptionKey)
+
+	}
+	rawGroupPrivateKey := e3dbClients.Base64Encode(groupKey[:])
+	fmt.Printf("Here is the response %+v  (same) \n", rawGroupPrivateKey)
```
galxy25, line 627: Can this debug log line be removed?

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+		ShareePublicKey:   queenClient.EncryptionKeys.Public.Material,
+	}
+	membershipKeyResponse, err := queenClient.CreateGroupMembershipKey(testCtx, membershipKeyRequest)
+	if err != nil {
+		t.Fatalf("Failed to create membership key \n response %+v \n error %+v", membershipKeyResponse, err)
+	}
+
+	// Add client to group
+	var CapabilityArray []string
+	CapabilityArray = append(CapabilityArray, ShareContentGroupCapability, ReadContentGroupCapability)
```
galxy25, line 688: FWIW you can rewrite these two lines above more concisely as
```golang
CapabilityArray := []string{CapabilityArray, ShareContentGroupCapability, ReadContentGroupCapability}
```

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+		GroupMember{
+			ClientID:        uuid.MustParse(groupMemberToAdd.ClientID),
+			MembershipKey:   membershipKeyResponse,
+			CapabilityNames: CapabilityArray})
+
+	addMemberRequest := AddGroupMembersRequest{
+		GroupID:      response.GroupID,
+		GroupMembers: memberRequest,
+	}
+	AddToGroupResponse, err := queenClient.AddGroupMembers(testCtx, addMemberRequest)
```
galxy25, line 701: local function variables should be camelcase, `addToGroupResponse`

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+		EncryptedGroupKey: response.EncryptedGroupKey,
+		ShareePublicKey:   queenClient.EncryptionKeys.Public.Material,
+	}
+	membershipKeyResponse, err := queenClient.CreateGroupMembershipKey(testCtx, membershipKeyRequest)
+	if err != nil {
+		t.Fatalf("Failed to create membership key \n response %+v \n error %+v", membershipKeyResponse, err)
+	}
+
+	// Add client to group
+	var CapabilityArray []string
```
galxy25, line 687: technically this is a slice (which is different from an array in go), but in either case adding the type of the variable to the name of the variable in a statically typed language (where the compiler and your editor can tell you what the type of every variable is) is almost always redundant, suggest `capabilities` or `groupMemberCapabilities`

galxy25: since this slice of capabilities is used in almost all the following tests, suggest moving this up to a file level variable with the other ones at line 22

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+	AddToGroupResponse, err := queenClient.AddGroupMembers(testCtx, addMemberRequest)
+	if err != nil {
+		t.Fatalf("Failed to Add Group Member to Group: Request:  %+v Err: %+v", addMemberRequest, err)
+	}
+	for _, requestGroupMember := range addMemberRequest.GroupMembers {
+		var found bool
+		for _, responseGroupMember := range *AddToGroupResponse {
+			if responseGroupMember.ClientID == requestGroupMember.ClientID {
+				found = true
+			}
```
galxy25, line 710: break if found

storageClient/storageClient_test.go
```diff
@@ -554,3 +555,274 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+	}
+	AddToGroupResponse, err := queenClient.AddGroupMembers(testCtx, addMemberRequest)
+	if err != nil {
+		t.Fatalf("Failed to Add Group Member to Group: Request:  %+v Err: %+v", addMemberRequest, err)
+	}
+	for _, requestGroupMember := range addMemberRequest.GroupMembers {
+		var found bool
+		for _, responseGroupMember := range *AddToGroupResponse {
+			if responseGroupMember.ClientID == requestGroupMember.ClientID {
+				found = true
```
galxy25, line 821: break if found, otherwise you'll loop through more times than needed

galxy25 (approved):

storageClient/storageClient_test.go
rosasck: I added it at the end of the function

rosasck: "if rawGroupPrivateKey != encryptionKeyPair.Private.Material {
		t.Fatalf("Raw Private key does not match: raw %+v client's raw private key %+v", encryptionKeyPair.Private.Material, rawGroupPrivateKey)
	}"

storageClient/storageClient_test.go
```diff
@@ -19,9 +19,10 @@ import (
 )
 
 var (
-	cyclopsServiceHost = os.Getenv("TOZNY_CYCLOPS_SERVICE_HOST")
-	testCtx            = context.Background()
-	ClientServiceHost  = os.Getenv("CLIENT_SERVICE_HOST")
+	cyclopsServiceHost    = os.Getenv("TOZNY_CYCLOPS_SERVICE_HOST")
+	testCtx               = context.Background()
+	ClientServiceHost     = os.Getenv("CLIENT_SERVICE_HOST")
+	FakeEncryptedGroupKey = "SZDkQQvA6srSZBy9tdYMAtgEE86fUzgLBnxP5bpz3U3bKiTWFllHGzpVU4o_VNNe._-U2eZWO0sKIZw0djuUCfPnvQHo72guB"
```
galxy25, line 25: why

rosasck: my check requires the bytes to be 97, since thats how many bytes a group key is but the way i was currently testing adding a group the encrypted group pair doesn't have the same number of bytes so i needed a way to allow my tests to work without having to go back and create membership keys for everything

galxy25: Here's my take: solve the problem (tediousness of creating a membership key) now so future developers of this code (including yourself) completely and correctly now so its not someone else's problem in the future.
Create a e3db-clients-go client/helper/test method/function that can create a valid membership key, and then use that wherever in the test you need to create a valid key. Once you solve this problem for yourself, you will have then provided a good foundation for other developers both external and internal to be able to use our product without having to work around any rough edges. Self love = customer love.

rosasck: working on this now and will transfer to storage service!

rosasck:
Ready for Final Review

storageClient/storageClient_test.go
```diff
@@ -555,3 +583,285 @@ func TestListGroupsWithClientForClientWithGroupsReturnsSuccess(t *testing.T) {
...
+				break
+			}
+		}
+		if !found {
+			t.Fatalf("Failed to add correct Group Members: Response( %+v) \n", AddToGroupResponse)
+		}
+	}
+}
+
+func CreateGroupMembershipKey(queenClient StorageClient, groupMember StorageClient, encryptedGroupKey string) (string, error) {
```
galxy25, line 857: move this over to https://github.com/tozny/e3db-clients-go/blob/master/test/test.go

galxy25 (approved):

PR #144, opened 2020-12-09
adding encrypted group key for create group in Read Group function

rosasck:


galxy25 (approved):

willmichael (approved):
In a future PR @ storage I would rework the response to be a 400 since what's actually happening is the user isn't providing proper input

PR #146, opened 2020-12-11
Tests for initiate endpoint with accountClient

carrala:


accountClient/api.go
```diff
@@ -215,3 +217,26 @@ type InternalSigClientInfoResponse struct {
 	PublicKey  string    `json:"public_key"`
 	SigningKey string    `json:"signing_key"`
 }
+
+// InitUpdateEmailRequest wraps values needed to issue a request to start an email update
+type InitiateUpdateEmailRequest struct {
+	AccountID    string    `json:"account_id"`
+	CurrentEmail string    `json:"current_email"`
+	NewEmail     string    `json:"new_email"`
+	CreatedAt    time.Time `json:"created_at"`
```
willmichael, line 226: Should this created at field be specified by the service?

willmichael (approved):

accountClient/api.go
galxy25: Agree, this field doesn't need to be in this request struct, as well as the CoolOffEnd as those fields will be determined by the server.

accountClient/api.go
```diff
@@ -215,3 +217,26 @@ type InternalSigClientInfoResponse struct {
 	PublicKey  string    `json:"public_key"`
 	SigningKey string    `json:"signing_key"`
 }
+
+// InitUpdateEmailRequest wraps values needed to issue a request to start an email update
+type InitiateUpdateEmailRequest struct {
+	AccountID    string    `json:"account_id"`
```
galxy25, line 223: We should be able to determine the account id for any authenticated client server side (talk to @rosasck for some examples on that) so this value should either be optional for this request(in which case it can stay as is), or be removed from this request.

galxy25: though if you do keep it also update this type to be a uuid (@efabens I can't believe you let your mentee do this ;-) )

accountClient/api.go
```diff
@@ -215,3 +217,26 @@ type InternalSigClientInfoResponse struct {
...
+
+// InitUpdateEmailResponse wraps data returned from /profile/email endpoint
+type InitiateUpdateEmailResponse struct {
+	AccountID      uuid.UUID `json:"account_id"`
+	CurrentEmail   string    `json:"current_email"`
+	NewEmail       string    `json:"new_email"`
+	NewEmailHash   string    `json:"new_email_hash"`
+	Verified       bool      `json:"email_verfied"`
+	UpdateComplete bool      `json:"update_complete"`
+	ExpiresAt      time.Time `json:"update"`
```
galxy25, line 238: should this be `expires_at`

accountClient/accountClient_test.go
```diff
@@ -158,3 +160,58 @@ func TestInternalGetClientAccountReturnsClientsAccountId(t *testing.T) {
...
+		CreatedAt:    currentTime,
+		CoolOffEnd:   currentTime.Add(time.Hour * 24),
+	}
+	// Make request to post / initiate the  email update
+	response, err := updateClient.InitiateEmailUpdate(testCtx, newEmailReq)
+	if err != nil {
+		t.Fatalf("Failed to initiate email update \n Email Req: (%+v) \n error %+v", newEmailReq, err)
+	}
+	// Verify that the AccountID returned matches the client
+	if response.AccountID.String() != newEmailReq.AccountID {
```
galxy25, line 187: more meaningful to check that the new email matches the requested new email and that the current email is still the original email.

accountClient/accountClient_test.go
```diff
@@ -158,3 +160,58 @@ func TestInternalGetClientAccountReturnsClientsAccountId(t *testing.T) {
...
+		CoolOffEnd:   currentTime.Add(time.Hour * 24),
+	}
+	// Make request to post / initiate the email update & verify it's successful
+	_, err = updateClient.InitiateEmailUpdate(testCtx, newEmailReq)
+	if err != nil {
+		t.Fatalf("Failed to initiate email update \n Email Req: (%+v) \n error %+v", newEmailReq, err)
+	}
+	// Make request to post / initiate the same email update again & verify it fails
+	_, err = updateClient.InitiateEmailUpdate(testCtx, newEmailReq)
+	if err == nil {
```
galxy25, line 214: (please talk to @rosasck / look at some of the tests for storageClient) on how to verify that not only an error was returned but that that status code is the expected value (assuming [409](https://http.cat/409) for this type of error)

galxy25 (approved):
Approved pending resolution of comments

PR #151, opened 2020-12-17
get group members endpoint

rosasck:


storageClient/storageClient.go
```diff
@@ -134,6 +134,18 @@ func (c *StorageClient) DeleteGroupMembers(ctx context.Context, params DeleteGro
 	err = e3dbClients.MakeSignedServiceCall(ctx, request, c.SigningKeys, c.ClientID, nil)
 	return err
 }
+
+// GetGroupMembers returns the group members and capabilites based on the groupID
+func (c *StorageClient) GetGroupMembers(ctx context.Context, params GetGroupMembersRequest) (*[]GroupMember, error) {
+	var result *[]GroupMember
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String() + "/members"
+	request, err := e3dbClients.CreateRequest("POST", path, nil)
```
galxy25, line 142: GET?

storageClient/api.go
```diff
@@ -52,6 +52,9 @@ type DescribeGroupRequest struct {
 	GroupID uuid.UUID `json:"group_id"`
 }
 
+// GetGroupMembersRequest wraps the group id needed to look up group members
+type GetGroupMembersRequest DescribeGroupRequest
```
galxy25, line 56: now that I think about it, for both this and the api spec, this should be thought of as a list endpoint, because it returns more than one resource, Get/Describe are the verbs to use for fetching a single resource.

storageClient/storageClient.go
```diff
@@ -134,6 +134,18 @@ func (c *StorageClient) DeleteGroupMembers(ctx context.Context, params DeleteGro
 	err = e3dbClients.MakeSignedServiceCall(ctx, request, c.SigningKeys, c.ClientID, nil)
 	return err
 }
+
+// GetGroupMembers returns the group members and capabilites based on the groupID
+func (c *StorageClient) GetGroupMembers(ctx context.Context, params GetGroupMembersRequest) (*[]GroupMember, error) {
```
galxy25, line 139: ListGroupMembers

storageClient/api.go
galxy25: ListGroupMembersRequest

willmichael (approved):

PR #154, opened 2020-12-18
Feature/interceptors

lkwdwrd:
This updates our client go library, adding the ability to use interceptors in our clients. This is a concept seen in many popular request libraries (axios in Javascript, okhttp in Java, etc.). 

I'm not super happy with the naming, esepically but ran with it to avoid additional refactor time. The PR itself is _very_ formulaic, so reading through should be pretty quick.

The Auth http client has been swapped to use an interceptor instead of the client from the oauth2 package -- under the hood it uses the same methods, but made the interceptor logic more consistent.

Fixed up some reported lint errors while in files, so you see minor tweaks, spelling updates, etc.

Tests are all passing against platform after some rebasing to catch new code from yesterday.

### DIFF
--- a/request/request.go
+++ b/request/request.go
@@ -1,0 +1,43 @@
+package request
+
+import "net/http"
+
+// Requester is a partial interface which allows decoration of the http Client
+type Requester interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// RequesterFunc is a function with acts as a Requester
+type RequesterFunc func(*http.Request) (*http.Response, error)
+
+// Do calls f(r)
+func (f RequesterFunc) Do(r *http.Request) (*http.Response, error) {
+	return f(r)
+}
+
+// Interceptor is a function which decorates a Requester
+//
+// The interceptor can determine whether to pass the request down the line using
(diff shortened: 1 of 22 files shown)


galxy25 (approved):

request/request.go
```diff
@@ -0,0 +1,43 @@
...
+// Do calls f(r)
+func (f RequesterFunc) Do(r *http.Request) (*http.Response, error) {
+	return f(r)
+}
+
+// Interceptor is a function which decorates a Requester
+//
+// The interceptor can determine whether to pass the request down the line using
+// the Do method, or perform some kind of operation on it, before or after it is
+// sent. I could even short circuit the req, sending back a valid http Response,
```
efabens, line 22: One could

efabens (approved):
Is the plan to make the signed service calls just use a tsv1 interceptor as well? Or is that out of scope for this

lkwdwrd:
TSV1 - out of scope for this, but yes, I think it should be an interceptor. Most of the special requests methods should probably eventually get placed into interceptors. There's quite a bit of good refactoring we could do in here to make things easier to work and reason with, but I cut myself off to keep the scope of this update smaller.

PR #155, opened 2020-12-18
sharing with groups

rosasck:
i decided this needed to be done before the list shared records with group so ill leave those changes but wont push them until this is complete

### DIFF
--- a/storageClient/api.go
+++ b/storageClient/api.go
@@ -47,6 +47,27 @@ type Group struct {
 	EncryptedGroupKey string    `json:"encrypted_group_key"`
 }
 
+// GroupAccessKeyRequest wraps the values to get or create an access key
+type GroupAccessKeyRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// ShareGroupRecordRequest wraps the values of a record for a group
+type ShareGroupRecordRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// ShareGroupRecordResponse wraps the return values of sharing a record for a group
+type ShareGroupRecordResponse ShareGroupRecordRequest
+
 // DescribeGroupRequest wraps values used to look up Groups
 type DescribeGroupRequest struct {
 	GroupID uuid.UUID `json:"group_id"`
@@ -86,6 +107,11 @@ type GroupMember struct {
 
 // DeleteGroupMembersRequest  wraps the information of all members being removed from group provided.
 type DeleteGroupMembersRequest AddGroupMembersRequest
+
+// GetMembershipKeyRequest wraps the information to get the membership key for a client
+type GetMembershipKeyRequest struct {
+	GroupID uuid.UUID `json:"group_id"`
+}
 
 // AddGroupMembersRequest  wraps the information of all members being added.
 type AddGroupMembersRequest struct {
--- a/storageClient/storageClient.go
+++ b/storageClient/storageClient.go
@@ -213,6 +213,59 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context, 
 
 }
 
+// GetGroupMembershipKey returns the client's group information including the membership key.
+func (c *StorageClient) GetGroupMembershipKey(ctx context.Context, params GetMembershipKeyRequest) (*ClientGroup, error) {
+	var result *ClientGroup
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String() + "/membership_key"
+	req, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, &result)
+	return result, err
+}
+
+// CreateGroupAccessKey takes in information about a record type and wraps the access key for a group.
+func (c *StorageClient) CreateGroupAccessKey(ctx context.Context, params GroupAccessKeyRequest) (e3dbClients.SymmetricKey, string, error) {
+	var wrappedAccessKey string
+	var accessKey e3dbClients.SymmetricKey
+	// Decrypt Access Key
+	rawEncryptionKey, err := e3dbClients.DecodeSymmetricKey(c.EncryptionKeys.Private.Material)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
+	accessKey, err = e3dbClients.DecryptEAK(params.EncryptedAccessKey, c.EncryptionKeys.Public.Material, rawEncryptionKey)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
+	// Encrypt Access Key with Group's Public Key
+	encryptionKeys := e3dbClients.EncryptionKeys{
+		Public: e3dbClients.Key{
+			Type:     e3dbClients.DefaultEncryptionKeyType,
+			Material: params.PublicKey,
+		},
+		Private: c.EncryptionKeys.Private,
+	}
+	eak, eakN, err := e3dbClients.BoxEncryptToBase64(accessKey[:], encryptionKeys)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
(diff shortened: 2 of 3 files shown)


storageClient/api.go
```diff
@@ -47,6 +47,32 @@ type Group struct {
 	EncryptedGroupKey string    `json:"encrypted_group_key"`
 }
 
+// GroupAccessKeyRequest wraps the values to get or create an access key
+type GroupAccessKeyRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
```
galxy25, line 55: Why / I don’t think either of these keys need to be a part of this request struct

storageClient/api.go
```diff
@@ -87,6 +113,9 @@ type GroupMember struct {
 // DeleteGroupMembersRequest  wraps the information of all members being removed from group provided.
 type DeleteGroupMembersRequest AddGroupMembersRequest
 
+// GetMembershipKeyRequest wraps the information to get the membership key for a client
+type GetMembershipKeyRequest AddGroupMembersRequest
```
galxy25, line 117: Only type alias if the data is literally the same. I don’t think to get a membership key we need to send along a field for GroupMembers

storageClient/storageClient.go
```diff
@@ -213,6 +213,60 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
 
 }
 
+// GetGroupMembershipKey returns member information about the group
```
galxy25, line 216: Is this comment correct?

storageClient/storageClient.go
```diff
@@ -213,6 +213,60 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
...
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String() + "/membership_key"
+	req, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, &result)
+	return result, err
+}
+
+// CreateGroupAccessKey
```
galxy25, line 228: Please complete me

storageClient/storageClient.go
```diff
@@ -213,6 +213,60 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
...
+		return accessKey, wrappedAccessKey, err
+	}
+	// Encrypt Access Key with Group's Public Key
+	groupPubKey := params.PublicKey
+	encryptionKeys := e3dbClients.EncryptionKeys{
+		Public: e3dbClients.Key{
+			Type:     e3dbClients.DefaultEncryptionKeyType,
+			Material: groupPubKey,
+		},
+		Private: c.EncryptionKeys.Private,
```
galxy25, line 248: I think the key to wrap should be the access key for this record, not the clients private key. Keep in mind how powerful and sensitive a clients private key is, if that gets compromised all their data could be read, versus if only an access key for a record is leaked only records of that type are compromised.

storageClient/storageClient.go
```diff
@@ -213,6 +213,60 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
...
+	}
+	eak, eakN, err := e3dbClients.BoxEncryptToBase64(accessKey[:], encryptionKeys)
```
galxy25, line 250: You can do better than me. Lines 250 - 254 should be extracted to a re-usable function that can be used both here and from the original function CreateSharingAccessKey this is based off of.

galxy25 (changes requested):
See comments

storageClient/api.go
```diff
@@ -47,6 +47,32 @@ type Group struct {
...
+type GroupAccessKeyRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// GroupRecord wraps the values of a record for a group
+type GroupRecord struct {
```
lkwdwrd, line 60: When is this returned? I hesitate to have a specific per-record struct like this, as opposed to a wrapping struct like when searching records. I worry this may cause type problems down the line.

lkwdwrd (commented):
Before approval I want to make sure we're solid on our Types. There are a lot of concepts both existing and getting added here, and getting the types right now will let us continue moving quickly later instead of refactoring. Just making a list of types we should pay attention to:

- Record (Accessed via read or search)
- QueryRecord (returned from query/search endoints. see https://github.com/tozny/api-specs/blob/master/storage/v1.swagger.yml#L1383)
- EncryptedAccessKey (Returned from the access key endpoint, or with search results)
  - This is extended for this feature, optionally having wrapping data with membership key info
- GroupMembershipKey (Returned form /memebership endpoint? contains encrypted private key)
- Group (Returned from group list endpoint, contains public key)

Am I missing some types? Feels like I am.

storageClient/api.go
rosasck: After you share a record with a group

rosasck:
Pushed changes

storageClient/api.go
galxy25: Yup, and based on our talk this should change to using pds.ListedRecord and expanding the GetEAKResponse to include a slice of wrapping keys.

storageClient/storageClient.go
```diff
@@ -213,6 +213,60 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
...
+	rawEncryptionKey, err := e3dbClients.DecodeSymmetricKey(c.EncryptionKeys.Private.Material)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
+	accessKey, err = e3dbClients.DecryptEAK(params.EncryptedAccessKey, c.EncryptionKeys.Public.Material, rawEncryptionKey)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
+	// Encrypt Access Key with Group's Public Key
+	groupPubKey := params.PublicKey
```
galxy25, line 242: this should come from params

storageClient/api.go
```diff
@@ -47,6 +47,31 @@ type Group struct {
...
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// ListGroupRecords wraps the values to list Records shared with a group
+type ListGroupRecords struct {
+	NextToken    int64
```
galxy25, line 70: ListGroupRecordsRequest or (response) so its clear when this is used without having to search for  the code that uses it.

storageClient/api.go
```diff
@@ -47,6 +47,31 @@ type Group struct {
...
+	Max          int
```
galxy25, line 71: if this is for the response then max and next token need json structs

galxy25 (changes requested):
see comments

rosasck:
Changes updates

storageClient/api.go
```diff
@@ -47,6 +47,31 @@ type Group struct {
...
+// ShareGroupRecordRequest wraps the values of a record for a group
+type ShareGroupRecordRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// ListGroupRecords wraps the values to list Records shared with a group
```
galxy25, line 68: safer to delete this and avoid the merge conflict

storageClient/storageClient.go
```diff
@@ -213,6 +213,59 @@ func (c *StorageClient) EncryptMembershipKeyForGroupMember(ctx context.Context,
...
+	eak, eakN, err := e3dbClients.BoxEncryptToBase64(accessKey[:], encryptionKeys)
+	if err != nil {
+		return accessKey, wrappedAccessKey, err
+	}
+	wrappedAccessKey = fmt.Sprintf("%s.%s", eak, eakN)
+	return accessKey, wrappedAccessKey, nil
+}
+
+// ShareRecordWithGroup shares a record type for all group members in given group
+func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params ShareGroupRecordRequest) (*ShareGroupRecordRequest, error) {
```
galxy25, line 258: sigh (at myself because of all the pin balling I'm putting you through with my nit picks),but obviously we can't have a request type being returned as a response, however we can have a type ShareGroupRecordResponse that is a type alias to that request type

galxy25 (approved):

PR #156, opened 2020-12-23
list shared with groups

rosasck:
List shared with groups endpoint

### DIFF
--- a/storageClient/api.go
+++ b/storageClient/api.go
@@ -47,6 +47,32 @@ type Group struct {
 	EncryptedGroupKey string    `json:"encrypted_group_key"`
 }
 
+// GroupAccessKeyRequest wraps the values to get or create an access key
+type GroupAccessKeyRequest struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
+	EncryptedAccessKey string    `json:"encrypted_access_key"`
+	PublicKey          string    `json:"public_key"`
+}
+
+// GroupRecord wraps the values of a record for a group
+type GroupRecord struct {
+	GroupID            uuid.UUID `json:"group_id"`
+	RecordType         string    `json:"record_type"`
+	RecordID           string    `json:"record_id"`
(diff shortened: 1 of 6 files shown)


storageClient/api.go
```diff
@@ -73,6 +73,12 @@ type ListGroupRecords struct {
 	GroupRecords []GroupRecord `json:"group_records"`
 }
 
+// ListGroupRecordsRequest wraps values needed for the request to list all records shared with a group
```
galxy25, line 76: Is it all records or all record types?

storageClient/storageClient.go
```diff
@@ -267,6 +267,22 @@ func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params GroupRe
...
+func (c *StorageClient) GetSharedWithGroup(ctx context.Context, params ListGroupRecordsRequest) (*ListGroupRecords, error) {
+	var result *ListGroupRecords
+	path := c.Host + storageServiceBasePath + "/groups/share"
+	req, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	urlParams := req.URL.Query()
+	urlParams.Set("nextToken", strconv.Itoa(int(params.NextToken)))
+	urlParams.Set("max", strconv.Itoa(int(params.Max)))
```
galxy25, line 280: Think about what happens if this isn’t set by the caller, it would default to the 0 value for its type, integer, which would result in no group records being returned? Probably best to only set this value as a query param if it is non zero.

storageClient/storageClient.go
```diff
@@ -267,6 +267,22 @@ func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params GroupRe
 	return result, err
 }
 
+// GetSharedWithGroup shares a record type for all group members in given group
```
galxy25, line 270: Comment is incorrect

galxy25 (approved):
Approved pending resolution of comments

lkwdwrd (approved):
With max defaults updated, this looks good

rosasck:
Made a new branch and transfered info over.. i tried rebasing off of the other branch and got in a wierd state

PR #158, opened 2020-12-24
listing all shared records to a group

rosasck:
The other branch got out of state very badly so i transferred the changed to this clean branch. When all fails, ABORT

### DIFF
--- a/storageClient/api.go
+++ b/storageClient/api.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/tozny/e3db-clients-go/pdsClient"
 )
 
 // Note is the API-level note object that mirrors the note JSON objects.
@@ -111,6 +112,27 @@ type DeleteGroupMembersRequest AddGroupMembersRequest
 // GetMembershipKeyRequest wraps the information to get the membership key for a client
 type GetMembershipKeyRequest struct {
 	GroupID uuid.UUID `json:"group_id"`
+}
+
+// ListGroupRecordsRequest wraps values needed for the request to list all records shared with a group, can filter by writer ids
+type ListGroupRecordsRequest struct {
+	GroupID   uuid.UUID `json:"group_id"`
+	WriterIDs []string  `json:"writer_ids"`
+	NextToken string
+	Max       int
+}
+
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList []pdsClient.ListedRecord `json:"results"`
+	NextToken  string                   `json:"next_token"`
+}
+
+// AccessKeyWrapper holds the information needed to unwrap each layer of access keys
+type AccessKeyWrapper struct {
+	MembershipKey string    `json:"membership_key"`
+	PublicKey     string    `json:"public_key"`
+	AccessKeyID   uuid.UUID `json:"access_key_id"`
 }
 
 // AddGroupMembersRequest  wraps the information of all members being added.
--- a/storageClient/storageClient.go
+++ b/storageClient/storageClient.go
@@ -291,6 +291,27 @@ func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params ShareGr
 	return result, err
 }
 
+// GetSharedWithGroup returns a list of all records shared with a group
+func (c *StorageClient) GetSharedWithGroup(ctx context.Context, params ListGroupRecordsRequest) (*ListGroupRecordsResponse, error) {
+	var result *ListGroupRecordsResponse
+	path := c.Host + storageServiceBasePath + "/groups/" + params.GroupID.String() + "/share"
+	req, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	urlParams := req.URL.Query()
+	urlParams.Set("next_token", params.NextToken)
+	if params.Max != 0 {
+		urlParams.Set("max", strconv.Itoa(int(params.Max)))
+	}
+	for _, writer := range params.WriterIDs {
+		urlParams.Add("writer_ids", writer)
+	}
+	req.URL.RawQuery = urlParams.Encode()
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, &result)
+	return result, err
+}
+
 func (c *StorageClient) ReadNote(ctx context.Context, noteID string, eacpParams map[string]string) (*Note, error) {
 	var result *Note
 	path := c.Host + storageServiceBasePath + "/notes"
--- a/pdsClient/api.go
+++ b/pdsClient/api.go
@@ -2,6 +2,8 @@ package pdsClient
 
 import (
 	"time"
+
+	"github.com/google/uuid"
 )
 
 const BatchReadRecordLimit = 1000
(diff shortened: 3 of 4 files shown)


storageClient/api.go
```diff
@@ -113,6 +114,28 @@ type GetMembershipKeyRequest struct {
...
+	NextToken int64
+	Max       int
+}
+
+// ExecutorQueryResponse contains the high level information for the response sent to client by the executor.
+type ExecutorQueryResponse struct {
+	ResultList   []pdsClient.ListedRecord `json:"results"`
+	LastIndex    int                      `json:"last_index"`
+	TotalResults int64                    `json:"total_results"`
+	SearchID     string                   `json:"search_id"`
```
galxy25, line 129: Don’t just copy and paste, read each line you’re pasting and ask yourself do you need it / does it need to be updated. This isn’t needed

rosasck: Thought that had to do with elastic search

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+	GroupID   uuid.UUID `json:"group_id"`
+	NextToken int64
+	Max       int
+}
+
+// ExecutorQueryResponse contains the high level information for the response sent to client by the executor.
+type ExecutorQueryResponse struct {
+	ResultList   []pdsClient.ListedRecord `json:"results"`
+	LastIndex    int                      `json:"last_index"`
+	TotalResults int64                    `json:"total_results"`
```
galxy25, line 128: If we have this that means on the server side we need to do a query to count the total number of records shared,otherwise it needs to be removed

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
 
+// ListGroupRecordsRequest wraps values needed for the request to list all records shared with a group
+type ListGroupRecordsRequest struct {
+	GroupID   uuid.UUID `json:"group_id"`
+	NextToken int64
+	Max       int
+}
+
+// ExecutorQueryResponse contains the high level information for the response sent to client by the executor.
+type ExecutorQueryResponse struct {
```
galxy25, line 125: ListGroupRecordsResponse, and update the comment appropriately.

storageClient/api.go
rosasck: Do we want this behavior?

storageClient/storageClient.go
```diff
@@ -291,6 +291,24 @@ func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params ShareGr
...
+	req, err := e3dbClients.CreateRequest("GET", path, params)
+	if err != nil {
+		return result, err
+	}
+	urlParams := req.URL.Query()
+	urlParams.Set("nextToken", strconv.Itoa(int(params.NextToken)))
+	if params.Max != 0 {
+		urlParams.Set("max", strconv.Itoa(int(params.Max)))
+	}
+	req.URL.RawQuery = urlParams.Encode()
```
galxy25, line 307: We also need support for query param that is a list of writer_ids to filter this list call with

storageClient/storageClient.go
```diff
@@ -291,6 +291,24 @@ func (c *StorageClient) ShareRecordWithGroup(ctx context.Context, params ShareGr
...
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, &result)
+	return result, err
+}
+
 func (c *StorageClient) ReadNote(ctx context.Context, noteID string, eacpParams map[string]string) (*Note, error) {
 	var result *Note
 	path := c.Host + storageServiceBasePath + "/notes"
```
galxy25, line 314: Reminder to update the api spec

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList   []pdsClient.ListedRecord `json:"results"`
+	LastIndex    int                      `json:"last_index"`
+	TotalResults int64                    `json:"total_results"`
+	Wrap         Wrappers                 `json: “wrappers,omit empty”`
+}
+
+// Wrappers holds the information needed to unwrap each layer of access keys
+type Wrappers struct {
```
galxy25, line 133: `AccessKeyWrapper`

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+	NextToken int64
+	Max       int
+}
+
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList   []pdsClient.ListedRecord `json:"results"`
+	LastIndex    int                      `json:"last_index"`
+	TotalResults int64                    `json:"total_results"`
+	Wrap         Wrappers                 `json: “wrappers,omit empty”`
```
galxy25, line 129: AccessKeyWrappers *[]AccessKeyWrapper `json:”access_key_wrappers,omitempty”`

storageClient/api.go
galxy25: Want yes, need, well that depends on how much work it is. I think it can be done similarly to the query to look up the group records themself, but instead of select * from its select count(*) from, if thats the case yes we need it, otherwise we can defer for now.

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+	WriterIDs []string  `json:"writer_ids"`
+	NextToken int64
+	Max       int
+}
+
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList []pdsClient.ListedRecord `json:"results"`
+	LastIndex  int                      `json:"last_index"`
+	Wrap       Wrappers                 `json: “wrappers,omit empty”`
```
galxy25, line 129: `AccessKeyWrappers *[]AccessKeyWrapper json:”access_key_wrappers,omitempty”`

galxy25: “ AccessKeyWrappers” is a pointer to a a slice of zero or more AccessKeyWrapper items, and if this value is nil don’t encode it to json / include the field access_key_wrappers in the response

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+	GroupID   uuid.UUID `json:"group_id"`
+	WriterIDs []string  `json:"writer_ids"`
+	NextToken int64
+	Max       int
+}
+
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList        []pdsClient.ListedRecord `json:"results"`
+	LastIndex         int                      `json:"last_index"`
```
galxy25, line 128: It honestly doesn’t matter which way you choose, but use either LastIndex OR (exclusive or ;-) NextToken for the request and response

galxy25 (approved):
Approved pending tests

storageClient/api.go
```diff
@@ -113,6 +114,27 @@ type GetMembershipKeyRequest struct {
...
+	GroupID   uuid.UUID `json:"group_id"`
+	WriterIDs []string  `json:"writer_ids"`
+	NextToken int64
+	Max       int
+}
+
+// ListGroupRecordsResponse returns all the records shared with a group and the values needed to unwrap.
+type ListGroupRecordsResponse struct {
+	ResultList        []pdsClient.ListedRecord `json:"results"`
+	NextToken         int                      `json:"last_index"`
```
galxy25, line 128: This struct tag (and possibly the api spec need to be updated to `next_token`

galxy25 (approved):
Approved pending resolution of comments, and tests

lkwdwrd (approved):

rosasck:
Will add tests soon :)

PR #161, opened 2021-01-25
adding a description for group

rosasck:


storageClient/api.go
```diff
@@ -46,6 +47,7 @@ type Group struct {
 	CreatedAt         time.Time `json:"created_at"`
 	LastModified      time.Time `json:"last_modified"`
 	EncryptedGroupKey string    `json:"encrypted_group_key"`
+	GroupDescription  string    `json:"group_description"`
```
galxy25, line 50: just Description for the field name as well, as for the json tag reminder to update api-spec

rosasck: will fix now!

storageClient/api.go
```diff
@@ -34,6 +34,7 @@ type Note struct {
 type CreateGroupRequest struct {
 	Name              string `json:"group_name"`
 	PublicKey         string `json:"public_key"`
+	GroupDescription  string `json:"group_description"`
```
efabens, line 37: this should match the below tag. It looks like the api-spec has it as `description` at the moment

galxy25 (approved):

PR #167, opened 2021-03-01
changing path name for private realm info path 

rosasck:
I changed the path name after seeing Luke's comment! To keep the path names more standard!


identityClient/identityClient.go
```diff
@@ -679,7 +679,7 @@ func (c *E3dbIdentityClient) RealmInfo(ctx context.Context, realmName string) (*
 // PrivateRealmInfo fetches the private realm information based on realm name.
 func (c *E3dbIdentityClient) PrivateRealmInfo(ctx context.Context, realmName string) (*PrivateRealmInfo, error) {
 	var info *PrivateRealmInfo
-	path := c.Host + identityServiceBasePath + "/realm/info/" + realmName
+	path := c.Host + identityServiceBasePath + "/realm/" + realmName + "/info"
```
galxy25, line 682: you can't do this because cyclops doesn't support plugins on dynamic paths (in this case where realm name will be different each request), it has to be a static route prefix like realm/info

galxy25 (changes requested):
pr can be closed

PR #168, opened 2021-03-02
Plat633/search identities update

rosasck:
Updated Search Identities Endpoint


efabens (approved):

galxy25 (approved):
reminder to update api spec as well (though you're probably already doing that)

lkwdwrd (approved):

PR #175, opened 2021-04-14
Add identity and storage methods to support sdk-login

efabens:
- Adds identity flow methods
- Adds storage service methods for fetching notes by name
- Adds crypto methods for note reading and validation
- This does not currently have tests as the functionality to derive credential notes is in e3db-go (which will cause a circular dependency) we may want to shift that code here to allow for better testing

### DIFF
--- a/authClient/api.go
+++ b/authClient/api.go
@@ -2,12 +2,12 @@ package authClient
 
 // ValidateTokenRequest represents a valid request to the auth service's /validate endpoint.
 type ValidateTokenRequest struct {
-    Token    string `json:"token"`    //The token to validate
-    Internal bool   `json:"internal"` //Whether this token belongs to an internal client
+	Token    string `json:"token"`    // The token to validate
+	Internal bool   `json:"internal"` // Whether this token belongs to an internal client
 }
 
-// ValidateTokenRequest represents a response to calling auth service's /validate endpoint.
+// ValidateTokenResponse represents a response to calling auth service's /validate endpoint.
 type ValidateTokenResponse struct {
-    ClientId string `json:"client_id"` //The client associated with this token
-    Valid    bool   `json:"valid"`     //Whether the token was valid
+	ClientId string `json:"client_id"` // The client associated with this token
+	Valid    bool   `json:"valid"`     // Whether the token was valid
 }
--- a/crypto.go
+++ b/crypto.go
@@ -30,6 +30,7 @@ const (
 	NISTCryptographicMode    = "NIST"
 	SymmetricKeySize         = 32
 	SigningKeySize           = 64
+	PublicSigningKeySize     = 32
 	NonceSize                = 24
 	SaltSize                 = 16
 	AccountDerivationRounds  = 1000
@@ -42,12 +43,39 @@ const (
 	FILE_BLOCK_SIZE              = 65536
 )
 
+var (
+	PublicKeySizeError         = fmt.Errorf("the provided key was not %v bytes long", PublicSigningKeySize)
+	SignatureVerificationError = fmt.Errorf("the provided message could not be verified with the provided key")
+)
+
 // SymmetricKey is used for fast encryption of larger amounts of data
 // also referred to as the 'SecretKey'
 type SymmetricKey *[SymmetricKeySize]byte
 
 // SigningKey is used for fast signing and verifying properties of data
 type SigningKey *[SigningKeySize]byte
+
+// PublicSigningKey is used to verify a signatures
+type PublicSigningKey = *[PublicSigningKeySize]byte
+
+// PublicSigningKeyFromBytes returns a PublicSigningKey derived from a byte slice
+func PublicSigningKeyFromBytes(publicSigningKeyBytes []byte) (PublicSigningKey, error) {
+	if len(publicSigningKeyBytes) != PublicSigningKeySize {
+		return nil, fmt.Errorf("PublicSigningKey must be %d bytes. Provided slice was %d bytes", PublicSigningKeySize, len(publicSigningKeyBytes))
+	}
+	var psk *[PublicSigningKeySize]byte
+	copy(psk[:], publicSigningKeyBytes)
+	return psk, nil
+}
+
+// PublicSigningKeyFromEncodedString returns a PublicSigningKey derived from a base64 encoded string
+func PublicSigningKeyFromEncodedString(publicSigningKey string) (PublicSigningKey, error) {
+	pskBytes, err := Base64Decode(publicSigningKey)
+	if err != nil {
+		return nil, err
+	}
+	return PublicSigningKeyFromBytes(pskBytes)
+}
 
 // Nonce is a type that represents nonce randomly generated unique value that is used once in encrypting data.
 type Nonce *[NonceSize]byte
@@ -60,6 +88,12 @@ type Key struct {
 
 // Keys is a map of key type to key material
 type Keys map[string]string
+
+// DecryptedData is a map of arbitrary key value pairs where value has been decrypted
+type DecryptedData = map[string]string
+
+// EncryptedData is a map of arbitrary key value pairs where the value is in an encrypted, "dotted quad" format
+type EncryptedData = map[string]string
 
 // AsymmetricKeypair wraps a public and private key used for
 // asymmetric cryptographic operations.
@@ -132,9 +166,9 @@ func GenerateKeyPair() (EncryptionKeys, error) {
 
 // EncryptData encrypts a collection of data of string key and values using
 // Tozny v1 Record encryption, returning the encrypted data.
-func EncryptData(Data map[string]string, ak SymmetricKey) *map[string]string {
+func EncryptData(data map[string]string, ak SymmetricKey) *EncryptedData {
 	encryptedData := make(map[string]string)
-	for k, v := range Data {
+	for k, v := range data {
 		dk := RandomSymmetricKey()
 		ef, efN := SecretBoxEncryptToBase64([]byte(v), dk)
 		edk, edkN := SecretBoxEncryptToBase64(dk[:], ak)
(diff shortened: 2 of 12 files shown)


crypto.go
```diff
@@ -398,6 +425,33 @@ func Sign(message []byte, key SigningKey) []byte {
 	return bytes[:offset]
 }
 
+// Verify a signature of message using the provided public key, for messages signed with TFSP1;ED25519;BLAKE2B
+func Verify(signature string, message string, publicSigningKey []byte) (string, error) {
```
galxy25, line 429: lets use this type please for the public signing key https://github.com/tozny/e3db-clients-go/pull/175/files#diff-69c5c48c1da9bf0e59ac32c7d0c0089fee30329073ffcc86cf7200f224f44ecfR50

crypto.go
```diff
@@ -398,6 +425,33 @@ func Sign(message []byte, key SigningKey) []byte {
...
+	if len(publicSigningKey) != PublicSigningKeySize {
```
galxy25, line 430: that way we can move (but not remove the need for) this verification logic to the constructor function, and let the verify function focus on verifying

crypto.go
```diff
@@ -136,11 +143,28 @@ func EncryptData(Data map[string]string, ak SymmetricKey) *map[string]string {
...
 // DecryptData decrypts a collection of data of string key and values encrypted using Tozny v1 Record encryption,
 // returning the decrypted data and error (if any).
-func DecryptData(Data map[string]string, ak SymmetricKey) (*map[string]string, error) {
+func DecryptData(data map[string]string, ak SymmetricKey) (*map[string]string, error) {
+	return decryptDataWithProcessFunction(data, ak, func(key string, message string) (string, error) { return message, nil })
+}
+
+// decryptDataWithProcessFunction decrypts a map of key value pairs where values are encrypted using Tozny v1 Record encryption
+// after description a provided function is run against each key-value pair. The result of the function if err = nil is set as the value
+func decryptDataWithProcessFunction(data map[string]string, ak SymmetricKey, postprocess func(key string, value string) (string, error)) (*map[string]string, error) {
```
galxy25, line 165: Why is this not a public function?

crypto.go
```diff
@@ -423,6 +474,50 @@ func SignField(fieldName, fieldValue string, privateSigningKey SigningKey, salt
 	return signedFieldPrefix + fieldValue, nil
 }
 
+// hashString returns a base64 encoded Blake2b hash of the provided message
+func hashString(message string) (string, error) {
```
galxy25, line 478: Why is this not a public function?

identityClient/api.go
```diff
@@ -121,6 +122,70 @@ type RegisterIdentityResponse struct {
...
+}
+
+// IdentityLoginRedirectResponse is returned by the IdentityLoginRedirect and contains access tokens for
+// retrieving clients and sessions
+type IdentityLoginRedirectResponse struct {
+	AccessToken string    `json:"access_token"`
+	TokenType   string    `json:"token_type"`
+	Expiry      time.Time `json:"expiry"`
+}
+
```
galxy25, line 188: This is really nice to see get added!

crypto.go
efabens: I usually hide helper methods that I can't think of good third party use cases. But I can make it public

galxy25, line 165: `*map[string]string` this should get a type alias so the code is more readable

crypto.go
efabens: This causes a need to copy the slice later but the overhead is not massive

galxy25 (changes requested):
This is off to a great start,  requesting the test be added and any logic from  e3db-go extracted and refactored as needed.

crypto.go
efabens: That link doesn't work

efabens:
Later this week I will move all of the crypto back out of e3db-go into e3db-clients go so that testing can happen

identityClient/identityClient.go
```diff
@@ -562,6 +563,50 @@ func (c *E3dbIdentityClient) GetToznyHostedBrokerInfo(ctx context.Context) (*Toz
 	return toznyHostedBrokerInfo, err
 }
 
+// InitiateIdentityLogin begins the standard 3rd party login flow with TozID
+func (c *E3dbIdentityClient) InitiateIdentityLogin(ctx context.Context, loginRequest IdentityLoginRequest) (*map[string]interface{}, error) {
+	var resp *map[string]interface{}
```
rosasck, line 568: Does this use `InitialLoginResponse`

rosasck (approved):

identityClient/identityClient.go
efabens: It is supposed to.  @lkwdwrd in other sdks we have just treated this as a map that gets passed back. Is that because we can't guarantee its shape or is this safe to strongly type?

lkwdwrd: I am afraid I don't recall specifically -- It was most likely a 'lets get this working and come back and add the type once the requirements are fully understood' situation, but the type was never added. We should define the type for it.

authClient/api.go
```diff
@@ -2,12 +2,12 @@ package authClient
...
-    Internal bool   `json:"internal"` //Whether this token belongs to an internal client
+	Token    string `json:"token"`    //The token to validate
+	Internal bool   `json:"internal"` //Whether this token belongs to an internal client
 }
 
 // ValidateTokenRequest represents a response to calling auth service's /validate endpoint.
 type ValidateTokenResponse struct {
-    ClientId string `json:"client_id"` //The client associated with this token
-    Valid    bool   `json:"valid"`     //Whether the token was valid
+	ClientId string `json:"client_id"` //The client associated with this token
```
galxy25, line 11: you're go linter doesn't complain about the claustrophobic comments `// The client ...` ?

crypto.go
```diff
@@ -132,9 +163,9 @@ func GenerateKeyPair() (EncryptionKeys, error) {
 
 // EncryptData encrypts a collection of data of string key and values using
 // Tozny v1 Record encryption, returning the encrypted data.
-func EncryptData(Data map[string]string, ak SymmetricKey) *map[string]string {
+func EncryptData(data map[string]string, ak SymmetricKey) *map[string]string {
```
galxy25, line 166: can the return type be *DecryptedData ?

crypto.go
```diff
@@ -143,14 +174,31 @@ func EncryptData(Data map[string]string, ak SymmetricKey) *map[string]string {
...
 // DecryptData decrypts a collection of data of string key and values encrypted using Tozny v1 Record encryption,
 // returning the decrypted data and error (if any).
-func DecryptData(Data map[string]string, ak SymmetricKey) (*map[string]string, error) {
+func DecryptData(data map[string]string, ak SymmetricKey) (DecryptedData, error) {
+	return DecryptDataWithProcessFunction(data, ak, func(key string, message string) (string, error) { return message, nil })
+}
+
+// DecryptDataWithProcessFunction decrypts a map of key value pairs where values are encrypted using Tozny v1 Record encryption
+// after description a provided function is run against each key-value pair. The result of the function if err = nil is set as the value
+func DecryptDataWithProcessFunction(data map[string]string, ak SymmetricKey, postprocess func(key string, value string) (string, error)) (DecryptedData, error) {
```
galxy25, line 196: super sweet function

crypto.go
efabens: We probably want it to be EncryptedData yeah?

crypto.go
```diff
@@ -405,6 +456,30 @@ func Sign(message []byte, key SigningKey) []byte {
 	return bytes[:offset]
 }
 
+// Verify a signature of message using the provided public key, for messages signed with TFSP1;ED25519;BLAKE2B
+func Verify(signature string, message string, publicSigningKey PublicSigningKey) (string, error) {
+
```
galxy25, line 461: extra line

authClient/api.go
efabens: I started using goland so I wonder if it is different than the vscode one, can update

crypto.go
efabens: It's what the cool kids are doing

crypto.go
```diff
@@ -469,6 +585,51 @@ func DeriveSigningKey(seed []byte, salt []byte, iter int) (*[32]byte, *[64]byte)
 	return publicKey, privateKey
 }
 
+func DeriveCredentials(username string, password string, realmName string, nameSalt string) (string, EncryptionKeys, SigningKeys, error) {
```
galxy25, line 588: function comment, `DeriveIdentityCredentials`, part of me thinks this function should live in main.go but the other half isn't so sure

identityClient/api.go
```diff
@@ -121,6 +122,70 @@ type RegisterIdentityResponse struct {
...
+	Nonce         string `json:"nonce" schema:"nonce"`
+	ClientID      string `json:"client_id" schema:"client_id"`
+	ResponseType  string `json:"response_type" schema:"response_type"`
+	Scope         string `json:"scope" schema:"scope"`
+	RedirectURI   string `json:"redirect_uri" schema:"redirect_uri"`
+	ResponseMode  string `json:"response_mode" schema:"response_mode"`
+	State         string `json:"state" schema:"state"`
+	Username      string `json:"username" schema:"username"`
+	Target        string `json:"target" schema:"target"`
+	AuthSessionID string `json:"auth_session_id" schema:"auth_session_id"`
```
galxy25, line 145: loving the sweet sweet schema

crypto.go
efabens: This was the last thing I added, it used to live in e3db-go but wanted to move it to this repo. I am happy to put it in main. Main felt very network focused and while this is one layer up it still seemed more crypto than network 🤷🏻‍♂️

identityClient/identityClient.go
```diff
@@ -3,6 +3,7 @@ package identityClient
 import (
 	"context"
 	"fmt"
+	"github.com/gorilla/schema"
```
galxy25, line 6: nit: std lib packages should be grouped together, this could go done to lines 13-14

efabens: wtf goland

identityClient/identityClient.go
```diff
@@ -26,11 +27,13 @@ const (
 	roleMapperResourceName        = "role_mapping"
 	realmLoginPathPrefix          = "/auth/realms"
 	realmLoginPathPostfix         = "/protocol/openid-connect/token"
+	realmLoginAuthPathPostfix     = "/protocol/openid-connect/auth"
 	applicationMapperResourceName = "mapper"
 )
 
 var (
 	internalIdentityServiceBasePath = fmt.Sprintf("/internal%s", identityServiceBasePath)
+	encoder = schema.NewEncoder()
```
galxy25, line 36: Can we get a comment and intentional revealing name like `httpFormSchemaEncoder`?

galxy25 (approved):

galxy25:
Elegant and well commented work 👏

authClient/api.go
efabens: Interestingly it is indifferent to the two. If I run format it won't go to either it just sticks with what is there

PR #176, opened 2021-04-20
upload & download an encrypted file

carrala:
Includes a test for the whole process!

It turns out the compression field is required in the filemeta to write a file record, but it doesn't return an error. It looks like the creation succeeded until you try to get the file record back. I've added a check that adds a value for compression if it isn't provided before writing the file.


pdsClient/pdsClient.go
```diff
@@ -748,6 +748,17 @@ func (c *E3dbPDSClient) GetOrCreateAccessKey(ctx context.Context, params GetOrCr
 	return e3dbClients.DecryptEAK(encryptedAccessKey, accessKeyResponse.AuthorizerPublicKey.Curve25519, rawEncryptionKey)
 }
 
+func (c *E3dbPDSClient) GetFileRecord(ctx context.Context, pendingFileID uuid.UUID) (*Record, error) {
```
rosasck, line 751: comment please

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (c *StorageClient) UploadFile(url string, encryptedFileName string, checksum string, size int) (int, error) {
+	file, err := os.Open(encryptedFileName)
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+	req, err := http.NewRequest("PUT", url, file)
+	if err != nil {
+		return 0, nil
```
rosasck, line 18: do we want to return the error code?

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil || resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return resp.StatusCode, err
+	}
+	return 0, nil
+}
+
+func (c *StorageClient) DownloadFile(url string, encryptedFileName string) (string, error) {
+	file, err := os.Create(encryptedFileName)
+	if err != nil {
+		return "", err
```
rosasck, line 33: `DownloadFile: Error creating file` ?

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return "e3db-clients-go:DownloadFile: get failed", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Sprintf("e3db-clients-go:DownloadFile: status is %d", resp.StatusCode), nil
+	}
+	_, err = io.Copy(file, resp.Body)
+	if err != nil {
+		return "", err
```
rosasck, line 46: `DownloadFile: Error copying file body` ?

storageClient/storageClient.go
```diff
@@ -515,6 +515,9 @@ func (c *StorageClient) WriteRecord(ctx context.Context, params Record) (*Record
 func (c *StorageClient) WriteFile(ctx context.Context, params Record) (*PendingFileResponse, error) {
 	var result *PendingFileResponse
 	path := c.Host + storageServiceBasePath + "/files"
+	if params.Metadata.FileMeta.Compression == "" {
```
rosasck, line 518: Quick Question:
omitempty for the fileMeta is for the sever correct? so it wouldn't cause issues here

carrala: it is for the server, but if 'Compression' isn't included in the fileMeta then the request from WriteFile doesn't actually register the record as a file and causes an error. PDS doesn't actually return an error in this case, so i'm setting this to make sure that it has the required info & this problem won't occur.

storageClient/files.go
carrala: yep! will do!

pdsClient/pdsClient.go
```diff
@@ -748,6 +748,18 @@ func (c *E3dbPDSClient) GetOrCreateAccessKey(ctx context.Context, params GetOrCr
 	return e3dbClients.DecryptEAK(encryptedAccessKey, accessKeyResponse.AuthorizerPublicKey.Curve25519, rawEncryptionKey)
 }
 
+// GetFileRecord gets a record for the file with the specified pendingFileID and returns it.
+func (c *E3dbPDSClient) GetFileRecord(ctx context.Context, pendingFileID uuid.UUID) (*Record, error) {
```
galxy25, line 752: `GetPendingFileRecord` please

storageClient/storageClient_test.go
```diff
@@ -275,6 +276,161 @@ func TestWriteFileWithBigMetadataKey(t *testing.T) {
...
+		t.Fatalf("Could not read %s file: %+v", plaintextFileName, err)
+	}
+	decrypted, err := ioutil.ReadFile(decryptedFileName)
+	if err != nil {
+		t.Fatalf("Could not read %s file: %+v", decryptedFileName, err)
+	}
+	compare2 := bytes.Equal(ptxt, decrypted)
+	if !compare2 {
+		t.Fatalf("%s and %s file contents do not match", plaintextFileName, decryptedFileName)
+	}
```
galxy25, line 431: awesome sauce

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
+package storageClient
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+)
+
+func (c *StorageClient) UploadFile(url string, encryptedFileName string, checksum string, size int) (int, error) {
```
galxy25, line 10: comment, and please extract these parameters into a struct

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
...
+	req.Header.Set("Content-Type", "application/octet-stream")
+	req.ContentLength = int64(size)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil || resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return resp.StatusCode, err
+	}
+	return 0, nil
+}
+
+func (c *StorageClient) DownloadFile(url string, encryptedFileName string) (string, error) {
```
galxy25, line 30: comment, and please extract these parameters into a struct

storageClient/files.go
galxy25: please make these functions in a package called file, not methods for StorageClient.

storageClient/files.go
galxy25: please make these functions in a package called file, not methods for StorageClient.

storageClient/files.go
```diff
@@ -0,0 +1,49 @@
...
+	defer file.Close()
+	resp, err := http.Get(url)
+	if err != nil {
+		return "e3db-clients-go:DownloadFile: get failed", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Sprintf("e3db-clients-go:DownloadFile: status is %d", resp.StatusCode), nil
+	}
+	_, err = io.Copy(file, resp.Body)
```
galxy25, line 44: https://stackoverflow.com/questions/11692860/how-can-i-efficiently-download-a-large-file-using-go/33853856

some comment here to the effect that this is the efficient and memory safe way of downloading files would be helpful to future readers (speaking for myself who just had to google to make sure this would be performant ;-)

storageClient/storageClient.go
galxy25: a comment here would be useful, but we'll want this default safe logic only in the sdk/e3db-go, e3db-clients-go needs to be dumb/ have no implicit behavior

galxy25 (approved):
approved pending resolution of comments

storageClient/storageClient.go
galxy25: @efabens @carrala does Storage service return an error if compression isn't included? If not can you add a ticket for that so that whenever we get to enforcing files to go through storage v2 it will be safer for everyone.

pdsClient/pdsClient.go
carrala: This was a mistake with variable names -- it should be fileID instead of pendingFileID since the file has already been committed. I've updated the names.

storageClient/storageClient.go
efabens: That is a great idea. What is really odd is that pds doesn't check it at creation or verification time either. Just at fetch time. @carrala in the storage service endpoints we should validate during the initial post that file meta has all three fields present.

efabens (approved):

PR #178, opened 2021-05-05
Secret key sizes

carrala:
these key length constants are used to check that the key lengths in a client credential object are valid.

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -38,13 +38,12 @@ const (
 	// The UUIDv5 is derived from a fixed namespace UUIDv4
 	// "794253a4-310b-449d-9d8d-4575e8923f40" and a version string
 	// "TFSP1;ED25519;BLAKE2B"
-	ToznyFieldSignatureVersionV1 = "e7737e7c-1637-511e-8bab-93c4f3e26fd9" // UUIDv5 TFSP1;ED25519;BLAKE2B
-	FILE_VERSION                 = 3
-	FILE_BLOCK_SIZE              = 65536
-	PublicSigningKeyLength       = 43
-	PrivateSigningKeyLength      = 86
-	PublicEncryptionKeyLength    = 43
-	PrivateEncryptionKeyLength   = 43
+	ToznyFieldSignatureVersionV1         = "e7737e7c-1637-511e-8bab-93c4f3e26fd9" // UUIDv5 TFSP1;ED25519;BLAKE2B
+	FILE_VERSION                         = 3
+	FILE_BLOCK_SIZE                      = 65536
+	Base64EncodedPublicSigningKeyLength  = 43
+	Base64EncodedPrivateSigningKeyLength = 86
+	Base64EncodedSymmetricKeyLength      = 43
 )
 
 var (
--- a/crypto.go
+++ b/crypto.go
@@ -41,6 +41,10 @@ const (
 	ToznyFieldSignatureVersionV1 = "e7737e7c-1637-511e-8bab-93c4f3e26fd9" // UUIDv5 TFSP1;ED25519;BLAKE2B
 	FILE_VERSION                 = 3
 	FILE_BLOCK_SIZE              = 65536
+	PublicSigningKeyLength       = 43
+	PrivateSigningKeyLength      = 86
+	PublicEncryptionKeyLength    = 43
+	PrivateEncryptionKeyLength   = 43
 )
 
 var (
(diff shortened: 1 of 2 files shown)


rosasck (approved):

efabens (changes requested):
I think that these are for base64 encoded length, which doesn't feel like the appropriate way to validate a key length, and so I would hesitate to have them be constants.

galxy25:
> I think that these are for base64 encoded length, which doesn't feel like the appropriate way to validate a key length, and so I would hesitate to have them be constants.

@efabens would you recommend that the go SDK should decode a credential secret (which btw would be nice to have such a function to convert from dashboard json string to go sdk creds 🤔 ) from string into the bytes and use the existing type to validate? The current approach follows the logic of the js-sdk where we aren't doing such a conversion.

@carrala at the very least we need to resolve the "why are these lengths different then the raw byte length", e.g. if these are base64 encoded version (which I believe they are), then the names should be  like `Base64EncodedPublicSigningKeyLength`. It's obnoxious (what is this, a java spring class name?) but less so if you think about the amount of confusion that has been generated by this ambiguity just in the past week, can you imagine how much longer it would take future developers to figure that out?

efabens:
Yes, that was the recommendation that I put here https://github.com/tozny/e3db-go/pull/57#discussion_r626905035
I would agree it is not what we do for the js-SDK. There is some additional computation cost it does mitigate any risk of padding issues that might come up, though we have strict rules on that and so rejecting based on base64 encoded in that case might be better as we want to ensure padding is correct.

carrala:
> Yes, that was the recommendation that I put here [tozny/e3db-go#57 (comment)](https://github.com/tozny/e3db-go/pull/57#discussion_r626905035)
> I would agree it is not what we do for the js-SDK. There is some additional computation cost it does mitigate any risk of padding issues that might come up, though we have strict rules on that and so rejecting based on base64 encoded in that case might be better as we want to ensure padding is correct.
> 
> I am not strongly opinionated about converting to bytes vs renaming the constant. @galxy25 @carrala
> 
> As a note we do have a way to go from dashboard json to go sdk creds in that we can json unmarshal and there is a method to take that struct I believe

This does look like a nicer, less hack-y way to check if the credential format is correct. In that case I won't need these new keys. If that works, I'll close this PR & add the changes to the e3db-go PR. @efabens @galxy25

galxy25:
> Yes, that was the recommendation that I put here [tozny/e3db-go#57 (comment)](https://github.com/tozny/e3db-go/pull/57#discussion_r626905035)
> I would agree it is not what we do for the js-SDK. There is some additional computation cost it does mitigate any risk of padding issues that might come up, though we have strict rules on that and so rejecting based on base64 encoded in that case might be better as we want to ensure padding is correct.
> 
> I am not strongly opinionated about converting to bytes vs renaming the constant. @galxy25 @carrala
> 
> As a note we do have a way to go from dashboard json to go sdk creds in that we can json unmarshal and there is a method to take that struct I believe

@carrala @efabens I'm fine with either approach (with a slight bias to the current approach with minor re-naming since it currently works and is compatible with how the Identity portal works). Note that for the byte based approach a plain json unmarshal will give you the base64 string value of the key and not the raw bytes, we'll want reverse versions of this sort of method https://github.com/tozny/e3db-clients-go/blob/master/crypto.go#L62 to do byte length validation.

efabens (approved):

crypto.go
```diff
@@ -38,9 +38,12 @@ const (
 	// The UUIDv5 is derived from a fixed namespace UUIDv4
 	// "794253a4-310b-449d-9d8d-4575e8923f40" and a version string
 	// "TFSP1;ED25519;BLAKE2B"
-	ToznyFieldSignatureVersionV1 = "e7737e7c-1637-511e-8bab-93c4f3e26fd9" // UUIDv5 TFSP1;ED25519;BLAKE2B
-	FILE_VERSION                 = 3
-	FILE_BLOCK_SIZE              = 65536
+	ToznyFieldSignatureVersionV1         = "e7737e7c-1637-511e-8bab-93c4f3e26fd9" // UUIDv5 TFSP1;ED25519;BLAKE2B
+	FILE_VERSION                         = 3
+	FILE_BLOCK_SIZE                      = 65536
```
galxy25, line 43: now in this repo, since there is some more consistency than in e3db-go, please use CamelCase for constants

galxy25 (approved):

PR #181, opened 2021-05-10
Update file names

carrala:
Uses tempFiles instead of hard-coded file names to avoid any conflicts with other files with these names


galxy25 (approved):

efabens (approved):
Pending temp file as we discussed

rosasck (approved):

PR #182, opened 2021-06-07
Plat675/realm settings update

rosasck:
Realm Settings Update Client Side + Happy Path Tests

### DIFF
--- a/identityClient/api.go
+++ b/identityClient/api.go
@@ -1,8 +1,9 @@
 package identityClient
 
 import (
-	"github.com/google/uuid"
 	"time"
+
+	"github.com/google/uuid"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


identityClient/api.go
```diff
@@ -814,3 +817,10 @@ type SearchIdentitiesInformation struct {
 	UserID        string    `json:"user_id"`
 	ClientID      uuid.UUID `json:"client_id"`
 }
+
+// RealmSettingsUpdateRequest wraps the setting available for realm admins to update
+type RealmSettingsUpdateRequest struct {
+	SecretsEnabled      bool     `json:"secrets_enabled"`
+	MFAAvailable        []string `json:"mfa_available"`
+	EmailLookupsEnabled bool     `sql:"email_lookups_enabled"`
```
galxy25, line 825: How can the server discriminate between no value provided by the user and the default value that go will give to this struct when deserialized from json?

rosasck:
Updating

PR #183, opened 2021-06-07
Realm Settings Update 

rosasck:
Client Side for Realm Settings Update + happy path tests


galxy25 (approved):

rosasck:
updated with queen path

PR #196, opened 2021-08-23
[PLAT718] Update Logger

rosasck:


keycloakClient/api.go
```diff
@@ -24,11 +26,19 @@ type Client struct {
 	tokens           map[string]*TokenInfo
 }
 
+// LoggingClient
+type LoggingClient struct {
+	stdClient http.Client
```
galxy25, line 31: hey hey, if I give you pseudo code, I don't wanna see it committed, you should know that CTO code isn't production level code ;-) Prefer `standardClient` or `defaultClient`, but none of this ambiguous/abbreviated variable naming stuff

rosasck: I got too excited, i am commenting things right now

keycloakClient/keycloakClient.go
```diff
@@ -62,13 +62,38 @@ func New(config Config) (*Client, error) {
 	httpClient := http.Client{
 		Timeout: config.Timeout,
 	}
+	// Check if logging is enabled, if it is, pass in logger
+	if config.EnabledLogging {
+		loggingClient := LoggingClient{
+			standardClient:   httpClient,
+			StructuredLogger: config.Logger,
+		}
+		return &Client{
```
galxy25, line 71: don't return from the if, it makes the code harder to read. Create a pointer variable for Client and assign to it inside the if

galxy25 (approved):

PR #197, opened 2021-08-23
[PLAT718] Const Types Transfer and auto refresh tokens

rosasck:
Added types that were still being pulled from keycloak client! and the forgotten function of auto refresh tokens

Officially fully switched over

### DIFF
--- a/keycloakClient/keycloakClient.go
+++ b/keycloakClient/keycloakClient.go
@@ -17,29 +17,35 @@ import (
 )
 
 const (
-	realmRootPath                   = "/auth/admin/realms"
-	userExtensionPath               = "/auth/realms"
-	clientResourceName              = "clients"
-	roleResourceName                = "roles"
-	roleByIDResourceName            = "roles-by-id"
-	groupResourceName               = "groups"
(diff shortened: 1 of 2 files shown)


carrala (approved):

keycloakClient/keycloakClient.go
```diff
@@ -79,6 +85,28 @@ func New(config Config) (*Client, error) {
...
+func (c *Client) AutoRefreshToken(realm string, username string, password string, onFailure func(error)) {
+	info, err := c.GetTokenInfo(realm, username, password, true)
+	if err != nil {
+		// Unable to fetch the token, allow userland to determine the correct
+		// behavior here -- retry, panic, log, etc...
+		onFailure(err)
+		return
+	}
+	// Refresh 5 seconds before the auth token expires
+	nextRefresh := info.Expires.Sub(time.Now().Add(5 * time.Second))
```
galxy25, line 99: you can ignore me if its too annoying, but can we have this come from the client config, default it to 5 if it's not set, and have identity service when setting up the client use a value from an environment variable to set it for our uses

galxy25 (approved):

PR #206, opened 2021-09-03
[PLAT-731] token races

pirtleshell:
I was digging deeply into a race condition that was causing identitiy management tests to fail, which is the same race condition that causes us to need retry logic on the identity tests in this repo.

Changes to this repo:
* i made the mutex wrap the entire `GetTokenInfo` func. 
  * in the best case where it finds the token in the `tokens` map, it makes three extra boolean checks before releasing the lock. in the worst case, it refreshes & replaces the token. The lock should still be held as a race condition exists in which it locks for read, some other goroutine grabs the token, this code replaces the token (invalidating the old), and then the other goroutine is attempting to use a no longer valid one.
  * as will be seen in my identity management PR, if you grab the token as a string, there is always a race condition in which it could be replaced before you use it. but this reduces that timeframe & provides extra protection.

* autorefreshing tokens that are refreshed will continue to autorefresh.
  * when we refresh a token that has a `refresher` (the func that actually managed the timed refreshing) we never create a new refresher for the new token. this mean that our autorefreshing tokens really only autorefreshed one time before just becoming non-refreshing tokens. i added a `autorefreshes` flag to TokenInfo that will allow previously autorefreshing tokens to be renewed with a new refresher.

* finally, a bug! The `tokens` map used `realm + username` as a key which allows for collisions. example: user `pirtle` at realm `magic` stores his token in the same place as user `irtle` at realm `magicp`. i converted the key to be a struct to avoid the collisions.

As a bonus, I am able to run the identity client test suite and no retires happen 🎉

### DIFF
--- a/keycloakClient/keycloakClient.go
+++ b/keycloakClient/keycloakClient.go
@@ -145,10 +145,12 @@ func (c *Client) GetTokenInfo(realm string, username string, password string, fo
 	var newTokenInfo *TokenInfo
 	var err error
 	key := realm + username
-	KeycloakTokenInfoLock.Lock()
+
 	// Get exclusive access to the token
+	KeycloakTokenInfoLock.Lock()
+	defer KeycloakTokenInfoLock.Unlock()
+
 	tokenInfo, exists := c.tokens[key]
-	KeycloakTokenInfoLock.Unlock()
 	if !exists || time.Now().After(tokenInfo.RefreshExpires) {
 		// If the token doesn't exist or can no longer be refreshed, get a new token
 		newTokenInfo, err = c.FetchToken(realm, username, password)
--- a/keycloakClient/keycloakClient.go
+++ b/keycloakClient/keycloakClient.go
@@ -94,6 +94,21 @@ func New(config Config) (*Client, error) {
 	return &client, nil
 }
 
+// tokenAutoRefresher returns a token refresher func that will automatically refresh the token
+// before the given expiration time
+func (c *Client) tokenAutoRefresher(expires time.Time, realm string, username string, password string, onFailure func(error)) *time.Timer {
+	// Refresh before the auth token expires
+	nextRefresh := expires.Sub(time.Now().Add(time.Duration(c.refreshAuthTokenBeforeExpiration) * time.Second))
+	refresher := time.AfterFunc(nextRefresh, func(realm, username, password string, onFailure func(error), c *Client) func() {
+		// send back a function which will re-call this method after the timeout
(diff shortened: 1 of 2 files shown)


pirtleshell:
[PLAT-731][WASH-1015]

keycloakClient/keycloakClient.go
```diff
@@ -104,16 +119,10 @@ func (c *Client) AutoRefreshToken(realm string, username string, password string
...
+	info.autorefreshes = true
+	info.onRefreshFailure = onFailure
 	// Pass in arguments to allow original args to get garbage collected.
-	info.refresher = time.AfterFunc(nextRefresh, func(realm, username, password string, onFailure func(error), c *Client) func() {
-		// send back a function which will re-call this method after the timeout
-		// capturing the arguments in a closure.
-		return func() {
-			c.AutoRefreshToken(realm, username, password, onFailure)
-		}
-	}(realm, username, password, onFailure, c))
```
pirtleshell, line 116: i extracted this to a first order func that could also be used by `GetTokenInfo` which is responsible for creating a new token on refresh.

keycloakClient/keycloakClient.go
```diff
@@ -135,11 +144,16 @@ func (t *tokenJSON) toTokenInfo(iat time.Time) *TokenInfo {
...
-	key := realm + username
-	KeycloakTokenInfoLock.Lock()
+	key := tokenMapKey{
+		realm:    realm,
+		username: username,
+	}
+
 	// Get exclusive access to the token
+	KeycloakTokenInfoLock.Lock()
+	defer KeycloakTokenInfoLock.Unlock()
```
pirtleshell, line 154: The mutex wraps the whole refreshing function now

keycloakClient/keycloakClient.go
```diff
@@ -135,11 +144,16 @@ func (t *tokenJSON) toTokenInfo(iat time.Time) *TokenInfo {
 func (c *Client) GetTokenInfo(realm string, username string, password string, force bool) (*TokenInfo, error) {
 	var newTokenInfo *TokenInfo
 	var err error
-	key := realm + username
```
pirtleshell, line 138: here is the original key that has been replaced with a struct that will prevent exceptionally clever or (un)lucky users from colliding and possibly getting a token they shouldn't have.

rosasck (approved):
🎊

Mrhea (approved):
🥳

house9 (approved):

PR #216, opened 2021-09-30
[WASH-1049] Add login note and pass reset tests

Mrhea:
These are tests from storage service that do not hit internal endpoints. They have been ran against dev and both pass.

### DIFF
--- a/storageClient/storageClient_test.go
+++ b/storageClient/storageClient_test.go
@@ -21,19 +21,21 @@ import (
 	"github.com/tozny/e3db-clients-go/file"
 	"github.com/tozny/e3db-clients-go/identityClient"
 	"github.com/tozny/e3db-clients-go/pdsClient"
+	"github.com/tozny/e3db-clients-go/storageClient"
 	storageClientV2 "github.com/tozny/e3db-clients-go/storageClient"
 	"github.com/tozny/e3db-clients-go/test"
 	"github.com/tozny/utils-go"
 )
 
 var (
-	cyclopsServiceHost = os.Getenv("TOZNY_CYCLOPS_SERVICE_HOST")
-	testCtx            = context.Background()
-	ClientServiceHost  = os.Getenv("CLIENT_SERVICE_HOST")
-	plaintextFileName  = "plaintext"
-	encryptedFileName  = "encrypted"
-	decryptedFileName  = "decrypted"
-	downloadedFileName = "downloaded"
+	cyclopsServiceHost   = os.Getenv("TOZNY_CYCLOPS_SERVICE_HOST")
+	testCtx              = context.Background()
+	ClientServiceHost    = os.Getenv("CLIENT_SERVICE_HOST")
+	plaintextFileName    = "plaintext"
+	encryptedFileName    = "encrypted"
+	decryptedFileName    = "decrypted"
+	downloadedFileName   = "downloaded"
+	identityLoginRetries = 10
 )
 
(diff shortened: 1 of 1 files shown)


rosasck (approved):

galxy25 (approved):

storageClient/storageClient_test.go
```diff
@@ -2051,6 +2053,94 @@ func TestListGroupMembersReturnsSuccess(t *testing.T) {
...
+	eacpParam["tozid_eacp"] = "this is the wrong tozid_eacp"
+	for i := 0; i <= identityLoginRetries; i++ {
+		_, err = StorageClient.ReadNote(testCtx, noteWithNoteID.NoteID, nil)
+		if err == nil {
+			t.Fatalf("Expected read note to fail due to incorrect TozIDEACP for note %+v\n", noteToWrite)
+		}
+	}
+
+	// Account should become locked. Attempt to get read the note.
+	_, err = StorageClient.ReadNote(testCtx, noteWithNoteID.NoteID, nil)
```
pirtleshell, line 2105: is there a  way to distinguish that its failing due to a lock on the account rather than because it has the wrong EACP? maybe you could lookup the lock itself to make sure it exists?

storageClient/storageClient_test.go
```diff
@@ -2051,6 +2053,94 @@ func TestListGroupMembersReturnsSuccess(t *testing.T) {
...
+	}
+
+	noteWithNoteID, err := StorageClient.WriteNote(testCtx, *noteToWrite)
+	if err != nil {
+		t.Fatalf("unable to write note %+v, to storage servive %s\n", noteToWrite, err)
+	}
+
+	// Attempt to get the same note with a different TozIDEACP
+	eacpParam := make(map[string]string)
+	eacpParam["tozid_eacp"] = "this is the wrong tozid_eacp"
```
pirtleshell, line 2096: how is this being used in the ReadNote call in the loop below?

storageClient/storageClient_test.go
house9: Good call, I think the e3db-client will just pass back the response from the server: "Expectation Failed" or "Unauthorized" in this case

Maybe we should check it explicitly? 
- `if strings.Contains(fmt.Sprint(err), "Expectation Failed")` for lock check and
- `if strings.Contains(fmt.Sprint(err), "Unauthorized")` for 401

house9 (approved):

storageClient/storageClient_test.go
pirtleshell: the string check could work. checking the status code would be more effective. i probably don't have the context to know if you can do this, but @carrala did a status check in identity service [here](https://github.com/tozny/identity-management-service/pull/222/files#diff-d00f75cf1660f5695f7b02430161785c4c36f95d45e706282cf190be5109c154R7322-R7329),

if you change the loop code to check the error, be aware that _i think_ because it is `<=` you may get the locked error on the final iteration of the loop

pirtleshell: and by "you" i mean @Mrhea haha did not read your username 😂

house9: > checking the status code would be more effective.

yeah that makes way more sense, I'm not sure what info is available in the error object

galxy25: @team++ for the catch, @rosasck ++ for the robust assertion

PR #226, opened 2021-10-12
[WR-173] add ApproveAccessRequests to identity client

house9:


identityClient/api.go
```diff
@@ -923,6 +924,17 @@ type AccessRequestResponse struct {
 	RealmName             string               `json:"realm_name"`
 }
 
+// ApproveAccessRequestsRequest specifies a list of AccessRequests to approve
+type ApproveAccessRequestsRequest struct {
+	AccessRequestIDs []int64 `json:"access_request_ids"`
```
house9, line 929: Post demo we may want to change this to accept an object, so we can include notes?

galxy25: agreed

house9: comments now part of request payload

efabens (approved):

PR #229, opened 2021-10-13
[WR202] [WR203] Activate Federator Connection

rosasck:


identityClient/identityClient.go
```diff
@@ -1119,3 +1119,37 @@ func (c *E3dbIdentityClient) InitiateFederationConnection(ctx context.Context, p
...
+			if key == authHeader {
+				isAuthHeader = true
+				break
+			}
+		}
+		if isAuthHeader {
+			req.Header.Add(key, val)
+		}
+	}
+	return e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, nil)
```
galxy25, line 1154: not asking you to do this, but just to be aware that "ideally" we would make a new auth method that does the logic in Ines 1141 - 1152 and use that versus the signed service call handler.

Mrhea (approved):

galxy25 (approved):

PR #267, opened 2021-12-01
[WR-317] access policy plugin data

house9:
https://toznysecurity.atlassian.net/browse/WR-317

see: https://github.com/tozny/identity-management-service/pull/323


identityClient/api.go
```diff
@@ -1025,6 +1025,9 @@ type AccessPolicy struct {
 	ApprovalRoles                []Role `json:"approval_roles"`
 	RequiredApprovals            int    `json:"required_approvals"`
 	MaximumAccessDurationSeconds int    `json:"max_access_duration_seconds"`
+	PluginType                   string `json:"plugin_type"`
+	PluginID                     string `json:"plugin_id"`
+	PluginSource                 string `json:"plugin_source"`
```
galxy25, line 1030: nit: now that I think about it more, `PluginApprovalSource` seems more intention revealing to me (and less confusing without comments to help distinguish between Type/Source)

galxy25 (approved):

identityClient/api.go
efabens: `PluginProcessLocation` it is both approval and creation of requests. So maybe more broad than approval source. I am ending up with 5 word names. `PluginFlowManagementDataReferenceLocation` 🤣

efabens (approved):

identityClient/api.go
house9: > `PluginFlowManagementDataReferenceLocation` 🤣

😄  

I am fine with anything at this point, I did comment the attributes in identity service api just not here.
Only downside to `PluginApprovalSource` is it doesn't tell the entire story (creation), but at the same time its probably enough info to know what it is for.

are either of these any better or maybe more confusing? 
`PluginMPCFlowSource`, `PluginMPCProcessingSource`

galxy25: both `PluginMPCFlowSource`, `PluginMPCProcessingSource` introduce new concepts / names (Flow and Processing), so at least to me are more confusing then `Approval`. For TozID mpc requests, the approval source is a role, I'd argue for Jira the approval source is a board.

galxy25: after more thunking....`PluginMPCFlowSource` is a better name, and any confusion can be mitigated with a helpful comment

PR #268, opened 2021-12-03
Fix failing Federation tests

Mrhea:
These tests were outdated with respect to the tests in Identity Service. No password note is written for new Identities, so `GetFederatedIdentitiesForSync` fails to read the note from storage service when performing the password note check.

### DIFF
--- a/identityClient/identityClient_test.go
+++ b/identityClient/identityClient_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/google/uuid"
 	e3dbClients "github.com/tozny/e3db-clients-go"
 	"github.com/tozny/e3db-clients-go/accountClient"
+	"github.com/tozny/e3db-clients-go/storageClient"
 	"github.com/tozny/e3db-clients-go/test"
 	"github.com/tozny/utils-go"
 )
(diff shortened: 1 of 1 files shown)


galxy25 (approved):

efabens (approved):

carrala (approved):

identityClient/identityClient_test.go
```diff
@@ -40,7 +41,39 @@ var (
...
+		Data:                *encryptedData,
+		Plain:               map[string]string{"extra1": "not encrypted", "extra2": "more plain data"},
+		Signature:           "signed",
+		MaxViews:            5,
+	}, err
+}
+
+func createIdentityAndRegisterWithRealm(t *testing.T, realm *Realm, registrationToken string, clientConfig e3dbClients.ClientConfig) *RegisterIdentityResponse {
 	identityName := uuid.New().String()
 	identityEmail := identityName + "@email.com"
```
carrala, line 78: should this be updated to a tozny email?

Mrhea: I don't think these are testing any email functionality, but it's definitely a good idea to get them updated just in case. I will update them 👍

Mrhea: Actually turns out these were updated, my branch was just outdated

PR #270, opened 2021-12-06
Restructure identity client tests

Mrhea:
- Moves federation tests into its own file
- Adds test helpers file to be used across test files. I moved the helper methods from `identityClient_test.go` to this new file. 
- Updating internal login lock tests that are failing

Tests in `identityClient_internal_test.go` were failing before this restructuring: 

```
    identityClient_internal_test.go:134: InternalCreateIdentityLoginAudit Error: invalid character '<' looking for beginning of value with request params: {RealmDomain:testinternalidentitystatusidentityaccountisunlocked62877146940033567 Status:fail RequestType:test with username ClientID:00000000-0000-0000-0000-000000000000 UserID:00000000-0000-0000-0000-000000000000 Username:freud IPAddress: CreatedAt:0001-01-01 00:00:00 +0000 UTC}
```

I'm not sure if my priority should shift to fixing it, or continue focus on release critical EAI docs and deployment to prod. I believe this to be an issue with the tests themself rather than an issue with the login lock feature


Mrhea:
Don't mind the large number of changes, it's because I moved helper methods to another file and tests to another file. Code is the same.

galxy25 (approved):

efabens (approved):

galxy25:
@Mrhea since you don't need the internal tests for your release (and future developers won't need the multi-realm test for their release), please make a ticket for the failing test and it should get taken care of from either
- once EAI is shipped, before switching to your next task fixing this issue
- discussed at the next sprint planning meeting and assigned to the person with the best availability to handle it

Mrhea:
> @Mrhea since you don't need the internal tests for your release (and future developers won't need the multi-realm test for their release thanks to your work!), please make a ticket for the failing test and it should get taken care of from either
> 
> * once EAI is shipped, before switching to your next task fixing this issue
> * discussed at the next sprint planning meeting and assigned to the person with the best availability to handle it

Will do!

PR #278, opened 2022-02-14
[WR388] Update Credential Representation

rosasck:
The base branch is the feature branch, and here are the client side resources

### DIFF
--- a/identityClient/identityClient.go
+++ b/identityClient/identityClient.go
@@ -1414,13 +1414,26 @@ func (c *E3dbIdentityClient) InitiateWebAuthnChallenge(ctx context.Context, para
 }
 
 // RegisterMFADevice registers & persists an MFA device to an identity
-func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params RegisterMFADeviceRequest) error {
+func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params RegisterMFADeviceRequest) (IdentityCredentialInformation, error) {
+	var response IdentityCredentialInformation
 	path := fmt.Sprintf("%s%s/%s", c.Host, identityServiceBasePath, mfaResourceName)
 	request, err := e3dbClients.CreateRequest("POST", path, params)
 	if err != nil {
-		return err
+		return response, err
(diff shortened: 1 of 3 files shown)


identityClient/identityClient.go
```diff
@@ -1414,13 +1414,26 @@ func (c *E3dbIdentityClient) InitiateWebAuthnChallenge(ctx context.Context, para
 }
 
 // RegisterMFADevice registers & persists an MFA device to an identity
-func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params RegisterMFADeviceRequest) error {
+func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params RegisterMFADeviceRequest) (IdentityCredentialInformation, error) {
```
pirtleshell, line 1417: the plan is for register to return the whole device list for the user. the api i wrote in keycloak just registers & responds 204. i admit that my reason to do this was mostly to not have to figure out how to make the necessary database call in keycloak land to find the newly registered device. my plan was to just call `ListIdentitiesMFACredentials` for the user and return _all_ their credentials on registration

would accept a case for doing the work to only get the one. but returning all is simpler imo

pirtleshell (approved):
approved :shipit: ~~pending change of register response type to be `ListIdentitesMFADeviceResponse` or an argument for only returning the one that was created.~~

identityClient/identityClient.go
rosasck: @pirtleshell This is returning all the identities credential information! 
```
// IdentityCredentialInformation wraps a User's configured MFA credential data
type IdentityCredentialInformation struct {
	ToznyID            uuid.UUID               `json:"tozny_id"`
	UserID             string                  `json:"user_id"`
	TOTPDeviceList     []TOTPAuthenticator     `json:"totp_devices"`
	WebAuthnDeviceList []WebAuthnAuthenticator `json:"webauthn_devices"`
}
```

pirtleshell: 🤦 👏 disregard

Mrhea (approved):

PR #281, opened 2022-02-16
[WR-388] Add GetUserCredentials keycloak method

Mrhea:
Fixes a bug in which a reference to the response object was not being passed into `MakeSignedServiceCall` for `ListIdentitiesMFACredentials` and `RegisterMFADevice`. 

Adds a new Keycloak Client method to call Keycloak's `/credentials` endpoint, which returns a list populated with each credential the user has.

### DIFF
--- a/identityClient/identityClient.go
+++ b/identityClient/identityClient.go
@@ -1422,7 +1422,7 @@ func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params Regis
 		return response, err
 	}
 	request.Header.Add(toznySessionHeader, params.SessionToken)
-	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, request, c.SigningKeys, c.ClientID, response)
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, request, c.SigningKeys, c.ClientID, &response)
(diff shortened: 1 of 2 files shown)


rosasck (approved):

identityClient/identityClient.go
```diff
@@ -1422,7 +1422,7 @@ func (c *E3dbIdentityClient) RegisterMFADevice(ctx context.Context, params Regis
 		return response, err
 	}
 	request.Header.Add(toznySessionHeader, params.SessionToken)
-	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, request, c.SigningKeys, c.ClientID, response)
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, request, c.SigningKeys, c.ClientID, &response)
```
galxy25, line 1425: if you update the signature of `MakeSignedServiceCall` to take a pointer to an interface (*interface{}) does the code still compile, and does doing it the old/broken way result in a compile time error?

galxy25 (approved):

identityClient/identityClient.go
Mrhea: Both result in a compile time error

PR #283, opened 2022-02-22
[WR-387] change login context from string -> interface{}

pirtleshell:
though we haven't made use of them, the login context can be more than
just strings. it is especially helpful to be anything so that it can
be a list of items.

**this is a breaking change**, as previously used login context values must
now be cast to strings.

i'm open for debate on whether or not this change should happen in this way. my main justification is that we state this repo is for internal use and this makes all login flow funcs more powerful.

it is used by one handler in identity service & the login method of e3db-go. if we implemented this change by introducing another client method with the new return type, both these login flow uses would need updated to the new thing.

my primary motivation for this is that the webauthn login flow should support multiple authentication devices. part of the process is sending information about the identity's registered devices so that the hardware security device(s) know which public key ids should be used to sign the login challenge.

### DIFF
--- a/identityClient/api.go
+++ b/identityClient/api.go
@@ -171,19 +171,19 @@ type InitialLoginResponse struct {
 	NoteID                                 string `json:"note_id"`
 }
 
-// IdentitySessionRequestResponse is returned by the IdentitySesssionRequest. It contains data related to what additional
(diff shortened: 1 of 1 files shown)


galxy25 (approved):
As part of this update, please update the handler and e3db-go to work with the new field type

Mrhea (approved):

PR #288, opened 2022-03-10
improve dev UX of testing external endpoints

pirtleshell:
adds make commands for testing only the external client funcs of account & pds services

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+# Make commands for testing should have the following pattern:
+# `test-{service}` should run tests that use only external-facing client functions
+# `test-internal-{service}` should run all tests, including ones calling internal endpoints
+# Tests that make use of internal endpoints must be prefixed "TestInternal".
+
 # Import environment file
 include .env
 # Source all variables in environment file
@@ -10,6 +15,7 @@ $(error TOZNY_PLATFORM_DIR is not set)
 endif
 
 export GO111MODULE=on
+export NONINTERNAL_PATTERN='^Test[^(Internal)]'
 
 all: clean build up test down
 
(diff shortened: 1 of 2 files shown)


galxy25 (approved):
❤️

Makefile
```diff
@@ -67,14 +68,21 @@ test-federation-identity:
 test-storage:
 	go test -count=1 -v storageClient/storageClient_test.go
 
+# Run all Account-V1 Integration tests
+test-account:
```
rosasck, line 72: I would flip this, we are more often than not running external tests 
so make test-account the external and then test-account-internal, following the identity pattern

rosasck (approved):

Makefile
pirtleshell: i was thinking that if we're running internal tests we want to run all the tests vs only wanting external. what do you think?

Mrhea (approved):

Makefile
rosasck: hmmm we need a pattern then, 
either changing identity to follow the same or renaming 
"test-only-external-account"

galxy25: I agree with Katie but am not strongly opposed to the other way, it's easy to flip either way in the future based on more experience of using these commands

galxy25: consistency++

pirtleshell: `test-account-external` is added here. but i see ya! consistency with the existing identity makes the most sense. gonna flip what i have here so that `test-*` is implicitly external only. `test-internal-*` should run... all tests or _only_ internal? my preference is for all.

rosasck: not a huge opinion on that, could be all and if so changing identity internal to do the same

pirtleshell: updated w/ following comment:
```make
# Make commands for testing should have the following pattern:
# `test-{service}` should run tests that use only external-facing client functions
# `test-internal-{service}` should run all tests, including ones calling internal endpoints
# Tests that make use of internal endpoints must be prefixed "TestInternal".
```

pirtleshell:
quick thoughts: we need to merge this in to prioritize the integration tests in jenkins.

future pirtle:
* check if updating the keycloak client used by the manage realm button to expect a platform.local url resolves the manage realm button
* it may even be the case that new realm populates that field from this variable and that the button would have worked on new realms.

PR #292, opened 2022-04-12
[wr-425]

anyatozny:
Add 2 new fields to the RealmSettingsUpdateRequest for the forgot flow customization


identityClient/identityClient_test.go
```diff
@@ -3476,3 +3478,58 @@ func TestListCreatedAccessRequestPolicy(t *testing.T) {
...
+		*unmarshalledRealmUpdateSettingsRequest.ForgotPasswordCustomLink)
+	verifyFieldUnmarshalledCorrectly(t, "ForgotPasswordCustomText", *realmUpdateSettingsRequest.ForgotPasswordCustomText,
+		*unmarshalledRealmUpdateSettingsRequest.ForgotPasswordCustomText)
+	if !reflect.DeepEqual(*realmUpdateSettingsRequest.MFAAvailable, *unmarshalledRealmUpdateSettingsRequest.MFAAvailable) {
+		t.Fatalf("MFAAvailable has unmarshalled incorectly, expected: %s, actual: %s", *realmUpdateSettingsRequest.MFAAvailable,
+			*unmarshalledRealmUpdateSettingsRequest.MFAAvailable)
+	}
+}
+
+func verifyFieldUnmarshalledCorrectly(t *testing.T, fieldName string, expected interface{}, actual interface{}) {
```
galxy25, line 3531: Nice! This would be a good candidate function for moving here to allow it to be used by other test files in other service repositories -> https://github.com/tozny/utils-go/blob/master/test/test.go

or moving here https://github.com/tozny/e3db-clients-go/blob/master/test/test.go so that its available to use for any other client tests in this repo

galxy25 (approved):

identityClient/identityClient_test.go
anyatozny: Thanks! I wanted to find something like that. Let me move it there.

rosasck (approved):

rosasck:
Seems like your commits aren't signed, let me know if i can help!

PR #297, opened 2022-05-05
[PLAT-976] Client side endpoints

rosasck:
DRAFT STATE

### DIFF
--- a/secureComputeClient/api.go
+++ b/secureComputeClient/api.go
@@ -1,1 +1,78 @@
 package secureComputeClient
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// SubscriptionManagers wraps the Manager object for a subscription
+type SubscriptionManagers struct {
+	ToznyClientID uuid.UUID `json:"tozny_client_id,omitempty"`
+}
+
+// SubscriptionRequest wraps the request to subscribe to a computation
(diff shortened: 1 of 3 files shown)


secureComputeClient/api.go
```diff
@@ -1 +1,78 @@
...
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// SubscriptionManagers wraps the Manager object for a subscription
+type SubscriptionManagers struct {
+	ClientID uuid.UUID `json:"client_id,omitempty"`
```
Mrhea, line 11: Maybe `ToznyClientID`? Unless this is a separate ID from user or tozny

secureComputeClient/api.go
```diff
@@ -1 +1,78 @@
...
+}
+
+// SubscriptionRequest wraps the request to subscribe to a computation
+type SubscriptionRequest struct {
+	ClientID             uuid.UUID              `json:"client_id,omitempty"`
```
Mrhea, line 16: Same ? as above, could be something different though I am not 100% familiar

Mrhea (approved):

secureComputeClient/api.go
rosasck: fixing!

PR #327, opened 2023-04-27
Plat 1277 authenticator app flow

santhosh-dc:


santhosh-dc:
Have released v0.0.236-alpha, will continue testing with this, if things are good we can release 0.0.236

rosasck (approved):

PR #363, opened 2024-04-29
Delete client and records changes.

karthikpdc:
- These API wrappers are added as part delete user earlier.
- Changes made to add realm name to the request.


rosasck (approved):
To this day i still dont know what calls this api hahha so based on our findings u are not breaking anything

PR #380, opened 2025-01-16
Revert user delete enhancement changes

karthikpdc:
I had merged user delete enhancement API changes which should not be until all the dependent services got tested. So i have reverted it because it impacts IDMS.

This is a revert of this PR https://github.com/tozny/e3db-clients-go/pull/376


karthikpdc:
Closed this PR to raise against tozflow branch.

PR #382, opened 2025-06-17
Tozscrt 64 implement user storage limit

Jayanthi-moorthi:
Description:
1. When Realm created , we need to store the user's storage limit for each user via storage client using " WriteUserStorageLimit"
2. Deleting Realm and Deleting users in the realm , we need to delete the data in the user storage table. To do so,we are calling "DeleteRealmStorageLimit" and "DeleteUserStorageLimitForUser" end points.

To be more secure, we are not exposed those endpoints. We are calling from identity service as internal calls.

### DIFF
--- a/identityClient/identityClient.go
+++ b/identityClient/identityClient.go
@@ -1729,3 +1729,39 @@ func (c *E3dbIdentityClient) InitiateIdentityProviderLogin(ctx context.Context, 
 	err = e3dbClients.MakeRawServiceCall(c.requester, req, &resp)
 	return resp, err
 }
+
+// InternalGetClientInfo gets client information from identity service
+func (c *E3dbIdentityClient) InternalGetClientInfo(ctx context.Context, clientID string) (IdentityInfo, error) {
+	var response IdentityInfo
+	path := c.Host + internalIdentityServiceBasePath + "/clients/" + clientID
+	req, err := e3dbClients.CreateRequest("GET", path, nil)
+	if err != nil {
+		return response, err
+	}
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, &response)
+	return response, err
+}
+
+// InternalGetMPCGroupInfo gets MPC group information from identity service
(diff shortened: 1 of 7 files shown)


identityClient/identityClient.go
```diff
@@ -890,6 +890,18 @@ func (c *E3dbIdentityClient) UpdateIdentityAttributes(ctx context.Context, param
 	return e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, nil)
 }
 
+// UpdateIdentityAttributes updates requested identity's attributes
+func (c *E3dbIdentityClient) GetRealmAttributeById(ctx context.Context, realmName string) (*RealmAttribute, error) {
```
tozny-migal, line 894: This seems to be getting a specific attribute (the storage limit) rather than Realm Attributes in general. If this is accurate, can we update the function name to convey this? Something like GetRealmStorageLimitById

storageClient/storageClient.go
```diff
@@ -826,6 +826,45 @@ func (c *StorageClient) InternalGetNoteInfo(ctx context.Context, noteName string
...
+		return err
+	}
+	urlParams := req.URL.Query()
+	urlParams.Set("realmId", realmId)
+	err = e3dbClients.MakeSignedServiceCall(ctx, c.requester, req, c.SigningKeys, c.ClientID, nil)
+	return err
+}
+
+// PrimeByNoteName primes the note with the provided noteName
+func (c *StorageClient) DeleteRealmStorageLimit(ctx context.Context, realmId string) error {
```
tozny-migal, line 842: Is there a reason the DeleteRealmStorageLimit call is in storage client rather than identity?

tozny-migal (commented):
Have some questions, please look at comments.

identityClient/identityClient.go
Jayanthi-moorthi: Yes Migal. I will update the method name and comments

storageClient/storageClient.go
tozny-migal: Resolved the other one, just a reminder that this one is still unresolved.

PR #384, opened 2025-08-08
Return error message from downstream service.

karthikpdc:
Updated "ReturnRawServiceCall" method to return Keycloak error. Previously, when a downstream service return error, the response body was ignored(not read). 

Instead, it returns a error string with request URL and status text. 

Now, i have made a change to read response body and if it is not null then it will return the error response else fall back to the legacy error message.


tozny-migal (changes requested):

tozny-migal (approved):
LGTM, feel free to merge and create the final tag for this change that you bring into IDMS

